PR #1, opened 2023-03-15
docs: Add DNS documentation

moan0s:


moan0s:
There is a bit of a conflict between the documentation here and `examples/hosts` as there we only assume to have an IP whereas I assume that there is some kind of base domain.

PR #2, opened 2023-03-15
Garage

moan0s:


group_vars/mash_servers
```diff
@@ -291,3 +291,36 @@ uptime_kuma_container_labels_traefik_tls_certResolver: "{{ devture_traefik_certR
...
+garage_user_uid: "{{ mash_playbook_uid }}"
+garage_user_gid: "{{ mash_playbook_uid }}"
+
+garage_container_additional_networks: |
+  {{
+    ([mash_playbook_reverse_proxyable_services_additional_network] if mash_playbook_reverse_proxyable_services_additional_network else [])
+  }}
+
+garage_container_labels_traefik_enabled: "{{ mash_playbook_reverse_proxy_type in ['playbook-managed-traefik', 'other-traefik-container'] }}"
+garage_garage_container_labels_traefik_docker_network: "{{ mash_playbook_reverse_proxyable_services_additional_network }}"
```
spantaleev, line 317: `garage_garage`?

docs/services/garage.md
```diff
@@ -0,0 +1,104 @@
+# Miniflux
```
spantaleev, line 1:
```suggestion
# Garage
```

docs/services/garage.md
```diff
@@ -0,0 +1,104 @@
...
+Then you can create your first bucket with 
+```
+ansible-playbook -i inventory/hosts setup.yml --tags=create-bucket --extra-vars "bucket=<bucketname>"
+```
+and get an access key for it with
+```
+ansible-playbook -i inventory/hosts setup.yml --tags=create-bucket --extra-key "bucket=<bucketname> keyname=<keyname>"
+```
+
+arage is very versatile so make sure to check out the [documentation](https://garagehq.deuxfleurs.fr/documentation/quick-start/). You can always execute you own commands by ssh-ing to the server and running
```
spantaleev, line 100:
```suggestion
Garage is very versatile so make sure to check out the [documentation](https://garagehq.deuxfleurs.fr/documentation/quick-start/). You can always execute you own commands by ssh-ing to the server and running
```

docs/services/garage.md
```diff
@@ -0,0 +1,104 @@
...
+    data_path: "{{ garage_node1_base_path }}/data"
+    rpc_bind_port: 3911
+    node_addr: "garage-node1.example.com"
+  - name: "node2"
+    gateway: false
+    capacity: 3
+    metadata_path: "{{ garage_node2_base_path }}/metadata"
+    data_path: "{{ garage_node2_base_path }}/data"
+    rpc_bind_port: 3921
+    node_addr: "garage-node2.example.com"
```
spantaleev, line 74: I suppose this configuration is somewhat confusion, because it's a bit of a contrived example?

Is there a use-case where one would want to run both the gateway and 2 nodes on the same server?

I suppose it would be nice to see an example of how to host each node separately and then add them all together. Perhaps a layout like this would be nice:

- one server which is both a gateway and a node (or 2 nodes, for making it even more complicated)
- one other server which is just a node

And then seeing how these 2 separate servers are wired together and integrated into the gateway.

Do each node need to know about each other node? Or can each node be completely unaware of the others (configuration-wise) and just talk to the gateway to discover the other nodes?

moan0s: This is more or less my actual setup. The idea is that for multiple drives on one server you run multiple garage daemons, this is discussed here: https://git.deuxfleurs.fr/Deuxfleurs/garage/issues/218

But I think it would be nice to show in this documentation two separate servers that are wired together. Will do.

The nodes can connect by talking to the gateway (or any other node) as long as they share the RPC secret

moan0s:
Not planned any time so closing for now

PR #6, opened 2023-03-17
Prometheus Node Exporter

etkecc:


group_vars/mash_servers
```diff
@@ -507,6 +509,49 @@ peertube_systemd_required_services_list: |
...
+########################################################################
+#                                                                      #
+# prometheus_node_exporter                                             #
+#                                                                      #
+########################################################################
+
+prometheus_node_exporter_enabled: true
+prometheus_node_exporter_uid: "{{ mash_playbook_uid }}"
+prometheus_node_exporter_gid: "{{ mash_playbook_gid }}"
+prometheus_node_exporter_base_path: "{{ mash_playbook_base_path }}/node-exporter"
```
spantaleev, line 521:
```suggestion
prometheus_node_exporter_enabled: true

prometheus_node_exporter_identifier: "{{ mash_playbook_service_identifier_prefix }}prometheus-node-exporter"

prometheus_node_exporter_base_path: "{{ mash_playbook_base_path }}/prometheus-node-exporter"

prometheus_node_exporter_uid: "{{ mash_playbook_uid }}"
prometheus_node_exporter_gid: "{{ mash_playbook_gid }}"
```

I suggest something like this:

- adding the identifier prefix ( `mash-` by default)

- making the directory name match the full name (`/mash/node-exporter` -> `/mash/prometheus-node-exporter`) - when we have a bunch of other Prometheus-related services they would all be prefixed with `prometheus-` nicely

PR #15, opened 2023-03-24
Fix hubsite example

sagat79:
Fix examples:
- hubsite_service_list_additional
- hubsite_service_list_auto

Update Uptime Kuma description


spantaleev:
Looks good!

Additionally, I think we can improve the formatting of these service objects as well. Right now it's just one long line with many things (`{'name': ..., 'url': ...., ...}`). It's not very readable.

PR #16, opened 2023-03-24
feat: Add firezone VPN

moan0s:


docs/services/firezone.md
```diff
@@ -0,0 +1,23 @@
...
+firezone_hostname: example.org
+
+firezone_default_admin_email: "user@invalid.org"
+firezone_default_admin_password: "<securepassword>"
+
+# Generate this with `openssl rand -base64 32`
+firezone_database_encryption_key: "<secret>"
+```
+
+Use `ansible-playbook -i inventory/hosts setup.yml --tags=firezone-create-or-reset-admin` to create the configured
```
spantaleev, line 22: Maybe `just run-tags firezone-create-or-reset-admin` is better

docs/services/firezone.md
```diff
@@ -0,0 +1,23 @@
...
+admin account or reset the password to the password set in `vars.yml`.
```
spantaleev, line 23: It might be worth adding a **Networking** subsection in **Configuration** which tells people what ports will be opened and potentially need to be forwarded if they're using another firewall, similar to bcebaafb29a9b8cf314d1.


A **Usage** section (like the one we have in most other service docs) may be helpful too.

moan0s: I added both sections (but kept Usage short, as the upstream documentation is quite good I think)

moan0s:
I reopened an issue to support --user here: https://github.com/firezone/firezone/issues/1506

PR #22, opened 2023-03-25
Add redis support to nextcloud

nielscil:
I've made some changes to add redis support to nextcloud. I still got some comments/questions. Could you guys give some feedback?

### Open issues

- [x]  Change needs https://github.com/mother-of-all-self-hosting/ansible-role-nextcloud/pull/1 and update of `requirements.yml`.
- [x] I'm not sure if it should be enabled by default. Maybe you guys got an opinion on that? Redis docs claims: `Enabling the Redis database service will automatically wire all other services to use it`. So either the docs for redis need to be changed or the docs for nextcloud.

### Comments
- `nextcloud_redis_enabled` is a new variable made in `group_vars/mash_servers`. I'm not sure if it's the right place to put it because I'm not very into ansible yet.

### DIFF
--- a/docs/services/nextcloud.md
+++ b/docs/services/nextcloud.md
@@ -9,6 +9,7 @@ This service requires the following other services:
 
 - a [Postgres](postgres.md) database
 - a [Traefik](traefik.md) reverse-proxy server
+- a [Redis](redis.md) data-store (optional)
 
 
 ## Configuration
@@ -26,6 +27,9 @@ nextcloud_enabled: true
 
 nextcloud_hostname: mash.example.com
 nextcloud_path_prefix: /nextcloud
+
+# Enable redis (optional)
+#nextcloud_redis_enabled: true
 
 ########################################################################
 #                                                                      #
--- a/docs/services/nextcloud.md
+++ b/docs/services/nextcloud.md
@@ -9,7 +9,7 @@ This service requires the following other services:
 
 - a [Postgres](postgres.md) database
 - a [Traefik](traefik.md) reverse-proxy server
-- a [Redis](redis.md) data-store (optional)
+- a [Redis](redis.md) data-store (optional), installation details [below](#redis)
 
 
 ## Configuration
@@ -28,8 +28,7 @@ nextcloud_enabled: true
 nextcloud_hostname: mash.example.com
 nextcloud_path_prefix: /nextcloud
 
-# Enable redis (optional)
-#nextcloud_redis_enabled: true
+# Redis configuration, as described below
 
 ########################################################################
 #                                                                      #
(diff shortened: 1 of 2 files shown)


spantaleev:
1. Why do you need `nextcloud_redis_enabled` at all?
2. I've been working on a new role for supporting [NetBox](https://docs.netbox.dev/en/stable/) (not ready yet). NetBox also requires Redis - 2 instances even. Which made me wonder.. is it safe to pass the same Redis instance to each service? I guess not, for multiple reasons:
- services may step on each others' toes
- services may potentially read data owned by other services. While Postgres can be shared because it supports authentication, sharing the same Redis instance seems more problematic..

3. I don't think anything should be enabled by default. Integrating Nextcloud with Redis seems like an optional feature with dubious benefits for most deployments (or am I underestimating it?). Also, many people would like to avoid installing Redis, etc. The docs page (`docs/services/nextcloud.md`) should certainly mention Redis and tell people how to enable it and integrate it, but it shouldn't force it as a default.

-----

Because of the discussions in point 2 in this comment, I believe that installing a dedicated Redis instance for each service  (Nextcloud, NetBox, etc.) is preferrable. Reusing the same instance sounds like a bad practice. I believe our docs should tell people how to run a dedicated instance -- something that I'm still in the process of figuring out for NetBox (it won't be very pretty).

The problem is that this playbook can only call out onto the Redis role once and with one set of parameters (vars). Installing multiple Redis instances in a single run does not seem possible, so we need an alternative.

My idea was to ask people to create additional hosts, like `nextcloud.example.com-redis`, with a minimal `vars.yml` configuration which installs a dedicated Redis instance (`/mash/nextcloud-redis` and `mash-nextcloud-redis.service`, etc). Then, the main host (e.g. `nextcloud.example.com`) can reference this service by:

- adjusting `nextcloud_redis_host`
- adjusting the required systemd services for Nextcloud (not very easy to do yet.. we need to add an additional variable)
- adjusting the container networks that the Nextcloud container is connected to (not very easy to do yet.. we need to add an additional variable)

Unless you have better ideas for how to solve this difficult problem of "dedicated Redis instances", I suggest you wait for the NetBox support (likely coming tomorrow) and copy the ideas from there.

nielscil:
> 1. Why do you need `nextcloud_redis_enabled` at all?

I wanted to make it disabled by default even when the redis role is enabled. I guess when point 2 is available, and every service has it's own redis instance, this won't be an problem anymore.

> Unless you have better ideas for how to solve this difficult problem of "dedicated Redis instances", I suggest you wait for the NetBox support (likely coming tomorrow) and copy the ideas from there.

Yeah,  you're right that does sound like a problem. I've not enough knowledge of ansible to help you out here. I'll wait and copy  :)

spantaleev:
See [our Redis docs](https://github.com/mother-of-all-self-hosting/mash-playbook/blob/main/docs/services/redis.md) and [our PeerTube docs](https://github.com/mother-of-all-self-hosting/mash-playbook/blob/main/docs/services/peertube.md) or [our NetBox docs](https://github.com/mother-of-all-self-hosting/mash-playbook/blob/main/docs/services/netbox.md) for more information and examples on how to integrate with Redis.

For Nextcloud, we need something similar.

spantaleev:
https://github.com/mother-of-all-self-hosting/ansible-role-nextcloud/pull/1 has been merged and tagged.

The group vars changes are also in place in `main` of this playbook already.

Looks like this might be ready to merge then?

nielscil:
Still needs some testing and https://github.com/mother-of-all-self-hosting/ansible-role-nextcloud/pull/2 to comply to the new redis configuration. I try to test it this afternoon/ tomorrow :)

nielscil:
It can be merged!

PR #35, opened 2023-04-02
Add funkwhale

moan0s:
So just some comments why I did things the way I did them:

* The frontend container is based on nginx and it is not possible to run as non-root. The actual app in the container runs as user. Also readonly is not really possible (stopped after the 6th mount/tmpfs). The capabilities are dropped as far as possible.
* Postgres is automatically connected, redis isn't (as per playbook standard)
* The manual user creation is needed as django does not allow non-interactive creation of users... *sigh*

### DIFF
--- a/group_vars/mash_servers
+++ b/group_vars/mash_servers
@@ -710,12 +710,16 @@ funkwhale_database_port: "{{ '5432' if devture_postgres_enabled else '' }}"
 funkwhale_database_password: "{{ '%s' | format(mash_playbook_generic_secret_key) | password_hash('sha512', 'db.funkwhale', rounds=655555) | to_uuid }}"
 funkwhale_database_username: "{{ funkwhale_identifier }}"
 
-funkwhale_systemd_required_services_list: |
+funkwhale_api_systemd_required_services_list_auto: |
   {{
-    (['docker.service'])
-    +
     ([devture_postgres_identifier ~ '.service'] if devture_postgres_enabled and funkwhale_database_hostname == devture_postgres_identifier else [])
   }}
+
+funkwhale_frontend_systemd_required_services_list_auto: |
+  {{
+    ([devture_postgres_identifier ~ '.service'] if devture_postgres_enabled and funkwhale_database_hostname == devture_postgres_identifier else [])
+  }}
+
 
 funkwhale_api_container_additional_networks_auto: |
   {{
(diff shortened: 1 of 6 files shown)


moan0s:
Should have RTFM... There is an option to create the superuser non-interactivly which I'll implement in a few minutes https://docs.djangoproject.com/en/4.1/ref/django-admin/#createsuperuser

group_vars/mash_servers
```diff
@@ -676,6 +690,66 @@ focalboard_container_labels_traefik_tls_certResolver: "{{ devture_traefik_certRe
...
+funkwhale_database_port: "{{ '5432' if devture_postgres_enabled else '' }}"
+funkwhale_database_password: "{{ '%s' | format(mash_playbook_generic_secret_key) | password_hash('sha512', 'db.funkwhale', rounds=655555) | to_uuid }}"
+funkwhale_database_username: "{{ funkwhale_identifier }}"
+
+funkwhale_systemd_required_services_list: |
+  {{
+    (['docker.service'])
+    +
+    ([devture_postgres_identifier ~ '.service'] if devture_postgres_enabled and funkwhale_database_hostname == devture_postgres_identifier else [])
+  }}
```
spantaleev, line 718: I don't see `funkwhale_systemd_required_services_list` in the role.

There are these though:
- `funkwhale_api_systemd_required_services_list`
- `funkwhale_frontend_systemd_required_services_list`

I think it'd be better to use `_default`, `_auto` and `_custom` variables for each of these and to override the `_auto` variables here (`funkwhale_api_systemd_required_services_list_auto` and `funkwhale_frontend_systemd_required_services_list_auto` to inject Postgres and to not bother with `docker.service` which will be part of the `_default` set).

requirements.yml
```diff
@@ -50,6 +50,9 @@
 - src: git+https://github.com/mother-of-all-self-hosting/ansible-role-focalboard.git
   version: v7.8.0-0
   name: focalboard
+- src: git+https://github.com/mother-of-all-self-hosting/ansible-role-funkwhale.git
+  version: v1.3.0-rc5-3
```
spantaleev, line 54: We usually avoid "unstable" versions like this, but.. given that:

- this is rc5 and a final release is probably just around the corner
- this is not upgrading any existing users from a stable version to a RC (we don't have any funkwhale users yet)

.. so I suppose it's OK and a good way to proceed.

moan0s: Yes, the idea was that there were some changes in the docker deployment from 1.2 -> 1.3 and 1.3 seemed stable enough to use the rc. Next upgrades will be with the usual stable versions

group_vars/mash_servers
moan0s: Fixed :+1:

PR #58, opened 2023-05-15
Postgres exporter

moan0s:
I ran into some issues with my postgres db lately, as all connections where exceeded. Postgres exporter should help debug&monitor the database. I use a dedicated user to run the monitoring (as running with a superuser would be a security risk).


moan0s:
This is now ready to review and uses the existing role (sourced from the mother-of-all-self-hosting organization)

spantaleev:
Thanks! I've updated some docs related to this.

PR #69, opened 2023-06-01
Mobilizon & Postgis

moan0s:
This PR adds support for Mobilizon and PostGIS. Thanks to the recent update to `3.1.0` an important issue in upstream is solved and I can finally make the PR.

The roles:
*  [mother-of-all-self-hosting/ansible-role-mobilizon](https://github.com/mother-of-all-self-hosting/ansible-role-mobilizon)
*  [mother-of-all-self-hosting/ansible-role-postgis](https://github.com/mother-of-all-self-hosting/ansible-role-postgis)


docs/services/mobilizon.md
```diff
@@ -0,0 +1,58 @@
...
+postgis_enabled: true
+
+# Put a strong password below, generated with `pwgen -s 64 1` or in another way
+postgis_connection_password: ''
+
+########################################################################
+#                                                                      #
+# /postgis                                                             #
+#                                                                      #
+########################################################################
```
spantaleev, line 48: It would probably be better if all this exists only on the `services/postgis.md` page.

For the Mobilizon docs page, you can then list Postgis as a dependency and link ot `services/postgis.md` - just like we do with the other services which have dependencies. Example: https://github.com/mother-of-all-self-hosting/mash-playbook/blob/main/docs/services/authentik.md#dependencies

PR #76, opened 2023-06-28
Add Influxdb

moan0s:
This PR adds influxdb, hopefully telegraf will follow. But as an external telegraf can push metrics this can be useful already.


docs/services/influxdb.md
```diff
@@ -0,0 +1,42 @@
...
+[InfluxDB](https://www.influxdata.com/) is a self-hosted time-series database, that this playbook can install, powered by the [mother-of-all-self-hosting/ansible-role-influxdb](https://github.com/mother-of-all-self-hosting/ansible-role-influxdb) Ansible role.
+
+## Configuration
+
+To enable this service, add the following configuration to your `vars.yml` file and re-run the [installation](../installing.md) process:
+
+```yaml
+########################################################################
+#                                                                      #
+# influxdb                                                           #
```
spantaleev, line 12: Alignment issues

docs/services/influxdb.md
```diff
@@ -0,0 +1,42 @@
...
+
+#influxdb_init: true
+#influxdb_init_username: "USERNAME"
+#influxdb_init_password: "SUPERSECRETPASSWORD"
+#influxdb_init_org: "EXAMPLE_ORG"
+#influxdb_init_bucket: "SOMEBUCKET"
+
+########################################################################
+#                                                                      #
+# /influxdb                                                          #
```
spantaleev, line 33: Alignment issues

docs/services/influxdb.md
```diff
@@ -0,0 +1,42 @@
...
+
+
+# Advanced configuration
+# To bootstrap an initial user, bucket and organization you can uncomment and replace
+
+#influxdb_init: true
+#influxdb_init_username: "USERNAME"
+#influxdb_init_password: "SUPERSECRETPASSWORD"
+#influxdb_init_org: "EXAMPLE_ORG"
+#influxdb_init_bucket: "SOMEBUCKET"
```
spantaleev, line 29: If this is enabled:

1. Is it safe to leave it in `vars.yml` for subsequent runs? Or will it break?
2. If you change the credentials subsequently, will it create yet another user or does it only work the first time around?

We should probably put the answers to these questions both here and inside `default/main.yml`.

spantaleev (commented):
Minor feedback here.

Here's one additional comment about the role itself:

It's better if all container-related things (creating networks, pulling images) in https://github.com/mother-of-all-self-hosting/ansible-role-influxdb/blob/bcca06d65ee07216bf8180c83a0b75c3417d0f94/tasks/install.yml are one next to another. Right now, network creation is at the top, container image pulling is toward the bottom.

docs/services/influxdb.md
moan0s: 1. Yes, it's save
2. This is only used for initial setup and will not change on subsequent runs (specifically this setting has no effect if an existing boltdb is found)

spantaleev:
Looks great! :+1:

PR #78, opened 2023-07-03
Add mosquitto

moan0s:
Another, mor feature complete MQTT broker

### DIFF
--- a/docs/services/mosquitto.md
+++ b/docs/services/mosquitto.md
@@ -17,7 +17,7 @@ mosquitto_enabled: true
 
 ## Usage
 
-After installation, you can use `just run-tags mosquitto-add-user --extra-vars=username=<username> --extra-vars=password=<password>` to create a user. For the setting to take effect, you must restart the container. To do that you can use `just restart-group mosquitto`.
+After installation, you can use `just run-tags mosquitto-add-user --extra-vars=username=<username> --extra-vars=password=<password>` to create a user. For the setting to take effect, you must restart the container. To do that you can use `just start-group mosquitto`.
 
 You can then start to send and subscribe to MQTT topics. Use port 1883 and the servers IP or any domain you configured to point at this server.
 
--- a/docs/services/mosquitto.md
+++ b/docs/services/mosquitto.md
@@ -10,11 +10,14 @@ To enable this service, add the following configuration to your `vars.yml` file 
 
 ```yaml
 mosquitto_enabled: true
+
+# If you need to change the MQTT port you can uncomment and adjust
+# mosquitto_container_mqtt_host_bind_port: "1884"
 ```
 
 ## Usage
 
-After installation, you can use `just run-tags mosquitto-add-user --extra-vars=username=<username> --extra-vars=password=<password>"` to create a user. For the setting to take effect, you must restart the container.
+After installation, you can use `just run-tags mosquitto-add-user --extra-vars=username=<username> --extra-vars=password=<password>` to create a user. For the setting to take effect, you must restart the container. To do that you can use `just restart-group mosquitto`.
 
 You can then start to send and subscribe to MQTT topics. Use port 1883 and the servers IP or any domain you configured to point at this server.
 
--- a/docs/services/mosquitto.md
+++ b/docs/services/mosquitto.md
@@ -1,5 +1,4 @@
-# Mosquitto Ansible Role
-
+# Mosquitto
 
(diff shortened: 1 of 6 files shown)


requirements.yml
```diff
@@ -106,6 +106,9 @@
 - src: git+https://github.com/mother-of-all-self-hosting/ansible-role-mongodb.git
   version: v6.0.6-0
   name: mongodb
+- src: git+https://github.com/mother-of-all-self-hosting/ansible-role-mosquitto.git
```
spantaleev, line 109: The README for this role could be improved by:

- linking to the official upstream website (https://mosquitto.org/). It doesn't seem to be doing that right now
- fix the broken **official container image** link going to `https://hub.docker.com/_/mosquitto`. It should be `https://hub.docker.com/_/eclipse-mosquitto` instead.

----

I've also noticed a typo in the name of this variable: `mosquitto_allow_anonymus_access`

----

Also:

- the `mosquitto_container_http_port` variable seems to be defined but to not be used in the configuration file and in the systemd service file
- `mosquitto_container_http_port` is a bad name for Mosquitto's MQTT port. It's not HTTP. Perhaps `mosquitto_container_mqtt_port` would be a better name?
- `mosquitto_container_mqtt_host_bind_port` is hardcoded to `1883`, just like the other variable. It's better if they are kept in sync by defining on in terms of the other. `mosquitto_container_mqtt_host_bind_port: "{{ mosquitto_container_mqtt_port }}"` in `defaults/main.yml`
- users don't actually need to adjust the port inside the container, it's just `mosquitto_container_mqtt_host_bind_port` that's relevant.. So you may mention the `mosquitto_container_mqtt_host_bind_port` variable on the `mosquitto.md` documentation page somewhere


----

In `tasks/install.yml`, it's better if all container-related Ansible tasks are close together.

Right now, `community.docker.docker_network` is way at the top of the file, while `community.docker.docker_image` is toward the bottom.

To be consistent with the other roles, it's better to move container network creation toward the bottom as well.

group_vars/mash_servers
```diff
@@ -1766,7 +1768,33 @@ mongodb_managed_databases_auto: |
...
+mosquitto_systemd_required_services_list: |
+  {{
+    (['docker.service'])
+  }}
+
+
+
+########################################################################
+#                                                                      #
+# /mosquitto                                                          #
```
spantaleev, line 1795: Alignment issue

group_vars/mash_servers
```diff
@@ -1766,7 +1768,33 @@ mongodb_managed_databases_auto: |
 #                                                                      #
 ########################################################################
 
+########################################################################
+#                                                                      #
+# mosquitto                                                           #
```
spantaleev, line 1773: Alignment issue

docs/services/mosquitto.md
```diff
@@ -0,0 +1,23 @@
...
+
+To enable this service, add the following configuration to your `vars.yml` file and re-run the [installation](../installing.md) process:
+
+```yaml
+mosquitto_enabled: true
+```
+
+## Usage
+
+After installation, you can use `just run-tags mosquitto-add-user --extra-vars=username=<username> --extra-vars=password=<password>"` to create a user. For the setting to take effect, you must restart the container.
```
spantaleev, line 17: Seems like there's an unmatched `"` here.

"Restart the container" is not very helpful to more novice users. We'd better tell people how: `just restart-group mosquitto`

group_vars/mash_servers
```diff
@@ -1766,7 +1768,33 @@ mongodb_managed_databases_auto: |
...
+
+mosquitto_base_path: "{{ mash_playbook_base_path }}/{{ mash_playbook_service_base_directory_name_prefix }}mosquitto"
 
+mosquitto_uid: "{{ mash_playbook_uid }}"
+mosquitto_gid: "{{ mash_playbook_gid }}"
+
+mosquitto_systemd_required_services_list: |
+  {{
+    (['docker.service'])
+  }}
```
spantaleev, line 1789: This seems to match the default value from the role's `defaults/main.yml`, so it can be removed from here.

group_vars/mash_servers
```diff
@@ -1766,7 +1768,33 @@ mongodb_managed_databases_auto: |
...
+
+
+
```
spantaleev, line 1792: Too many empty lines. We only need 1 to be consistent with the other blocks.

PR #79, opened 2023-07-08
Add support for Frigate

cvwright:
This PR adds initial support for the Frigate network video recorder https://github.com/blakeblackshear/frigate

### DIFF
--- a/docs/services/frigate.md
+++ b/docs/services/frigate.md
@@ -1,0 +1,73 @@
+# Frigate
+
+[Frigate](https://frigate.video/) is an open source NVR built around real-time AI object detection.
+
+## Dependencies
+
+No other services are required to run Frigate in the default configuration,
+but it is commonly deployed together with an MQTT broker such as:
+* [mosquitto](mosquitto.md)
+* [rumqttd](rumqttd.md)
+
+## Configuration
+
+To enable this service, add the following to your `vars.yml` and re-run the [installation](../installing.md).
+
+```yaml
+########################################################################
+#                                                                      #
+# frigate                                                              #
+#                                                                      #
+########################################################################
+
+frigate_enabled: true
+
+# Required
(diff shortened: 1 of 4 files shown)


spantaleev:
Thanks for contributing support for Frigate!

Here's some feedback:

- it'd be nice to add a documentation page
- it'd be nice to expose it through Traefik like all other services. When done, `frigate_container_http_host_bind_port` should be unset. Traefik can also be used in local-only mode and without SSL certificates, so it shouldn't be preventing anyone from using Frigate in a local-only environment.
- environment variables are [defined inline in the systemd service file](https://github.com/cvwright/ansible-role-frigate/blob/c7a26fd23c557d2a86a6edb5f1a6137cc6502e2c/templates/systemd/frigate.service.j2#L35-L37). It's better to use an `env` file which is passed via `--env-file`. See [an example here](https://github.com/mother-of-all-self-hosting/ansible-role-authentik/blob/main/templates/env.j2). This has 4 benefits:
  - systemd will not need to be reloaded when the environment variables change (the playbook reloads it automatically anyway, so it's not that big of a deal)
  - any sensitive data in environment variables would not be visible from every user the machine doing `ps aux`
  - the systemd service file will be cleaner
  - adding a variable for extending the `env` file will be possible and easy
- in `tasks/install.yml`, it's better to keep related actions close to one another. The file currently starts with creating a Docker network, then does a bunch of file-related stuff, then goes back to pulling Docker images. Most roles do all Docker-related operations at the bottom of the file.
- `defaults/main.yml` mixes capitalization for booleans. It sometimes uses `false`, other times `False. It's better and more consisent with the rest of the playbook by using lowercase everywhere

cvwright:
Thanks @spantaleev!  It's great to get feedback from the expert.

The latest commits address most of your comments.  The big to-do item that remains is to add support for Traefik.

PR #82, opened 2023-07-19
Adds support for n8n

kinduff:
This PR adds support to n8n using a custom role I made at ([kinduff/ansible-docker-n8n](https://github.com/kinduff/ansible-docker-n8n)).

Works like a charm, I have an instance up and running in production in my main server (works in my server, lol).

There's a big caveat and I need an advice:

The n8n Docker image uses UID and GID 1000 for the user `node` internally. This cannot be remapped as it executes `chown` and `setgroups` commands. I tried a good amount of things but all are related to modifying or extending their existing Docker image which is not ideal. Are there any other roles in this playbook that had this problem?

Besides that, works good if you specify in your config the UID and GID n8n needs:

```yml
n8n_uid: "1000"
n8n_gid: "1000"
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,9 @@
 # ignore roles pulled by ansible-galaxy
 /roles/galaxy/*
 !/roles/galaxy/.gitkeep
+
+# ignores vscode file
+.vscode
+
+# ingores macos files
+.DS_Store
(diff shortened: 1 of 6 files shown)


spantaleev:
Awesome! Thank you for contributing n8n support!

I suppose if it can't be gotten to work with another uid/gid, we should hardcode `n8n_uid: "1000"` and `n8n_gid: "1000"` in the role's `defaults/main.yml` and put some comments there with the reason for doing it.

Besides this, I've taken a look and it seems good for merging!

PR #83, opened 2023-07-20
Adds support for Linkding

kinduff:
This PR adds support for Linkding using a custom role I made ([kinduff/ansible-docker-linkding](https://github.com/kinduff/ansible-docker-linkding)).

Works pretty good! Tested SQlite and PostgreSQL as engines, and with and without a path.

### DIFF
--- a/docs/services/linkding.md
+++ b/docs/services/linkding.md
@@ -6,7 +6,7 @@
 
 This service requires the following other services:
 
--   a [Postgres](postgres.md) database
+-   a [Postgres](postgres.md) database, but [SQLite](https://www.sqlite.org/) is also a possibility (see `linkding_database_engine` below)
 -   a [Traefik](traefik.md) reverse-proxy server
 
 ## Configuration
@@ -24,6 +24,12 @@ linkding_enabled: true
 
 linkding_hostname: mash.example.com
 linkding_path_prefix: /linkding
+
+# We configure Linkding to use Postgres by default. See docs/services/postgres.md.
+# To use an external Postgres server, you need to tweak additional `linkding_database_*` variables.
+# Feel free to remove the line below to make Linkding use SQLite.
+linkding_database_engine: postgres
+
 linkding_superuser_username: change me
 linkding_superuser_password: change me
 
--- a/docs/services/linkding.md
+++ b/docs/services/linkding.md
@@ -1,0 +1,53 @@
+# Linkding
+
+[Linkding](https://github.com/sissbruecker/linkding) bookmark manager that is designed be to be minimal and fast.
+
+## Dependencies
+
+This service requires the following other services:
+
+-   a [Postgres](postgres.md) database
+-   a [Traefik](traefik.md) reverse-proxy server
+
+## Configuration
+
+To enable this service, add the following configuration to your `vars.yml` file and re-run the [installation](../installing.md) process:
+
+```yaml
+########################################################################
+#                                                                      #
+# linkding                                                             #
+#                                                                      #
+########################################################################
+
+linkding_enabled: true
+
+linkding_hostname: mash.example.com
+linkding_path_prefix: /linkding
+linkding_superuser_username: change me
+linkding_superuser_password: change me
+
+########################################################################
+#                                                                      #
+# /linkding                                                            #
+#                                                                      #
+########################################################################
+```
+
+In the example configuration above, we configure the service to be hosted at `https://mash.example.com/linkding`.
+
+You can remove the `linkding_path_prefix` variable definition, so that the service is served at `https://mash.example.com/`.
+
+## Usage
+
+### Superuser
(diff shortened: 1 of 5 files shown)


docs/services/linkding.md
```diff
@@ -0,0 +1,53 @@
...
+
+Please note the use of [`linkding_superuser_username`](https://github.com/sissbruecker/linkding/blob/master/docs/Options.md#ld_superuser_name) and [`linkding_superuser_password`](https://github.com/sissbruecker/linkding/blob/master/docs/Options.md#ld_superuser_password) variables. These are not mandatory and are meant to be set the first time you run this role.
+
+### Database
+
+The role defaults to SQlite, but you can opt to use PostgreSQL by adding the following:
+
+```yaml
+linkding_database_engine: postgres
+```
```
spantaleev, line 53: This section seems better off as a subsection of the **Configuration** section, not under **Usage**.

I suppose the same goes for the **Superuser** section.

---

As for the database stuff, since it's quite important, I believe it's better to add something like this to the initial configuration example. See my other review comment and code-change suggestion.

docs/services/linkding.md
```diff
@@ -0,0 +1,53 @@
...
+# linkding                                                             #
+#                                                                      #
+########################################################################
+
+linkding_enabled: true
+
+linkding_hostname: mash.example.com
+linkding_path_prefix: /linkding
+linkding_superuser_username: change me
+linkding_superuser_password: change me
```
spantaleev, line 28:
```suggestion
linkding_enabled: true

linkding_hostname: mash.example.com
linkding_path_prefix: /linkding

# We configure Linkding to use Postgres by default. See docs/services/postgres.md.
# To use an external Postgres server, you need to tweak additional `linkding_database_*` variables.
# Feel free to remove the line below to make Linkding use SQLite.
linkding_database_engine: postgres

linkding_superuser_username: change me
linkding_superuser_password: change me
```

docs/services/linkding.md
```diff
@@ -0,0 +1,53 @@
+# Linkding
+
+[Linkding](https://github.com/sissbruecker/linkding) bookmark manager that is designed be to be minimal and fast.
+
+## Dependencies
+
+This service requires the following other services:
+
+-   a [Postgres](postgres.md) database
```
spantaleev, line 9:
```suggestion
-   a [Postgres](postgres.md) database, but [SQLite](https://www.sqlite.org/) is also a possibility (see `linkding_database_engine` below)
```

docs/supported-services.md
```diff
@@ -26,6 +26,7 @@
 | [Jitsi](https://jitsi.org/) | A fully encrypted, 100% Open Source video conferencing solution | [Link](services/jitsi.md) |
 | [Keycloak](https://www.keycloak.org/) | An open source identity and access management solution. | [Link](services/keycloak.md) |
 | [Lago](https://www.getlago.com/) | Open-source metering and usage-based billing | [Link](services/lago.md) |
+| [linkding](https://github.com/sissbruecker/linkding/) | Bookmark manager designed be to be minimal and fast. | [Link](services/linkding.md) |
```
spantaleev, line 29:
```suggestion
| [linkding](https://github.com/sissbruecker/linkding/) | Bookmark manager designed to be minimal and fast. | [Link](services/linkding.md) |
```

spantaleev (commented):
Thank you for contributing another cool service!

I'm posting some feedback about this PR.

docs/services/linkding.md
kinduff: Thanks for the view! The project defaults to SQLite, as well as the role. Shouldn't this implementation default to SQLite too? That's why I added it as a suggestion. What do you think?

spantaleev: I think that the role defaulting to SQLite is probably good, as it's less external dependencies.

Most components of this playbook are automatically hooked to its Postgres instance. Using a better database and having a convenient way to back up and restore all your databases seems useful.

Of course, we can make the playbook default to SQLite as well and ask people to uncomment a suggested `# linkding_database_engine: postgres` line in the main configuration. I think whichever way we do it, it's clear enough and people will find out early enough (as soon as they copy the configuration sample), so that it wouldn't be much trouble.

kinduff: Sounds about right! Applied your changes. Do you have other suggestions to make this clearer?

spantaleev: I think it looks good!

About the **Usage** section, we should remove this **Database** subsection as it's no longer needed. Also, the **Superuser** section seems more related to **Configuration** than to **Usage**. Maybe we can relocate that section there and remove the empty **Usage** section after that?

kinduff: Done! With this changes it makes way more sense, thanks a lot for the guidance!

PR #85, opened 2023-08-01
Add Changedetection.io, a simple website change detection and restock monitoring solution

nielscil:


spantaleev:
Thank you for contributing this!

I was thinking `browserless/chrome` could probably be installed via a separate role and this one could have it as a dependency, but.. we could extract it later on.

nielscil:
Yeah, thought the same when I finished the role. Gonna check if I can extract it when I've got the time

PR #87, opened 2023-08-15
Document adding alert rules

moan0s:
This is a PR as using this technique requires manual intervention on the server and we can question if wen want that to be an "offical" way of doing things. As I benefit from doing things like that I'd argue yes but I'm open to discussion

### DIFF
--- a/docs/services/prometheus.md
+++ b/docs/services/prometheus.md
@@ -86,6 +86,16 @@ prometheus_hostname: prometheus.example.com
 # prometheus_basic_auth_password: ''
 ```
 
+### Adding alert rules
+
+```
+# The alert rules were manually created in /mash/prometheus/config/alert.yml
+# any update must be made there
+# Setting this variable will tell prometheus to use the file you created
+prometheus_config_rule_files_additional:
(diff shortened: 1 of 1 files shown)


docs/services/prometheus.md
```diff
@@ -86,6 +86,16 @@ prometheus_hostname: prometheus.example.com
...
 ```
 
+### Adding alert rules
+
+```
+# The alert rules were manually created in /mash/prometheus/config/alert.yml
+# any update must be made there
+# Setting this variable will tell prometheus to use the file you created
+prometheus_config_rule_files_additional:
+  - "/etc/prometheus/alert.yml"
```
spantaleev, line 96: This example could also make use of [ansible-role-aux](https://github.com/mother-of-all-self-hosting/ansible-role-aux) role and manage the `/mash/prometheus/config/alert.yml` file using a `aux_file_definitions` variable.

Example:

```yaml
aux_file_definitions:
  - dest: "{{ prometheus_config_path }}/alert.yml"
    owner: "{{ prometheus_uid }}"
    group: "{{ prometheus_gid }}"
    content: |
      something
      another
```

PR #114, opened 2023-11-02
feat: Add semaphore

moan0s:
This is a basic support for ansible-semaphore. The service installs correctly and playbooks can be pulled. I hope to add some documentation on running a first playbook in the future but already wanted to add it as this might also be nice to be done by the community


moan0s:
Closed in favor of #125

PR #119, opened 2023-11-25
Adds Filebrowser

kinduff:
Adds [Filebrowser](https://filebrowser.org), a self-hosted file managing interface.

### DIFF
--- a/docs/services/filebrowser.md
+++ b/docs/services/filebrowser.md
@@ -1,0 +1,64 @@
+# Filebrowser
+
+[Filebrowser](https://filebrowser.org) is a self-hosted file managing interface.
+
+## Dependencies
+
+This service requires the following other services:
+
+- a [Traefik](traefik.md) reverse-proxy server
+
+## Configuration
+
+To enable this service, add the following configuration to your `vars.yml` file and re-run the [installation](../installing.md) process:
+
+```yaml
+########################################################################
+#                                                                      #
+# filebrowser                                                          #
+#                                                                      #
+########################################################################
+
+filebrowser_enabled: true
+
+filebrowser_hostname: filebrowser.example.com
+
+########################################################################
+#                                                                      #
+# /filebrowser                                                         #
+#                                                                      #
+########################################################################
+```
+
+
+## Usage
+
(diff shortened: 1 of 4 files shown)


moan0s:
Hi so I had a look and tested it. Worked out of box and the role looks good overall, thanks for adding this! 🎉

Some things I noticed (mixed regarding the role and this playbook if I should split it I can do that):
* Is there a reason we can't use the docker image to modify the configuration but use the binary?
* Please add the role to the readme (something like powered by [kinduff/ansible-docker-filebrowser](https://github.com/kinduff/ansible-docker-filebrowser.git)
* There doesn't  seem to be a reason to use the environment file. If you don't plan on using it you can remove it [here](https://github.com/kinduff/ansible-docker-filebrowser/blob/85752efa7456cb5091f648dbda2fe07faa833a19/templates/systemd/filebrowser.service.j2#L23), here [https://github.com/kinduff/ansible-docker-filebrowser/blob/85752efa7456cb5091f648dbda2fe07faa833a19/tasks/install.yml#L21] and the template itself
* As I see it you use [this task](https://github.com/kinduff/ansible-docker-filebrowser/blob/85752efa7456cb5091f648dbda2fe07faa833a19/tasks/install.yml#L24-L33) to create a database file when it doesn't exist. You could possibly simplify this with [ansible.builtin.file](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/file_module.html)
```yaml
- name: Ensure a file is present
  ansible.builtin.file:
    path: /path/foo
    state: touch
```
* You can add Filebrowser to the list of [supported services](https://github.com/mother-of-all-self-hosting/mash-playbook/blob/main/docs/supported-services.md)
* I tested it and Filebrowser seems to run fine with the option `--read-only` which you could set in [/templates/systemd/filebrowser.service.j2](https://github.com/kinduff/ansible-docker-filebrowser/blob/85752efa7456cb5091f648dbda2fe07faa833a19/templates/systemd/filebrowser.service.j2)
* The license in [meta/main.yml](https://github.com/kinduff/ansible-docker-filebrowser/blob/85752efa7456cb5091f648dbda2fe07faa833a19/meta/main.yml#L20) is GPL while the LICENSE file states AGPL

If you have any questions feel free to ask here or on matrix :)

PR #123, opened 2023-12-01
Add nextcloud_container_image_customizations_smb_enabled: true  to co…

brush:
…nfiguration options


brush:
updated this to work with the PR in nextcloud role

PR #125, opened 2023-12-06
feat: Add ansible-semaphore

moan0s:
Replaces #114 as the new playbook layout is too different.


docs/services/semaphore.md
```diff
@@ -0,0 +1,43 @@
...
+```yaml
+########################################################################
+#                                                                      #
+# semaphore                                                            #
+#                                                                      #
+########################################################################
+
+semaphore_enabled: true
+semaphore_hostname: semaphore.example.com
+semaphore_admin_password: STRONG_PASSWORD
```
spantaleev, line 26: Perhaps we should define this as empty and have a line recommended how to generate a strong password (`pwgen -s 64 1`). I believe we do this for other roles, so it'd be consistent.

I'd also add some new lines between the various setting groups, so that it's not a large wall-of-text. Example:

```yml
semaphore_enabled: true

semaphore_hostname: semaphore.example.com

# Despite the confusing naming, semaphore_admin_name needs to hold a username, not a name!
semaphore_admin_name: ''
semaphore_admin_email: ''
# You can generate a strong password with a command like: `pwgen -s 64 1`
semaphore_admin_password: ''

# Key for encrypting access keys in database.
# It must be generated by using the following command: head -c32 /dev/urandom | base64
semaphore_access_key_encryption: ''
```

It's unfortunate that `semaphore_admin_name` is poorly named (due to the environment variable for Semaphore being poorly named).

For `semaphore_access_key_encryption` (and other variables), it's probably better to leave them empty instead of showing some example value (which people may forget to change and keep as-is)? Empty values are supposedly a better fail-safe default, as `tasks/validate_config.yml` should catch and report them as undefined.

PR #129, opened 2023-12-13
bump nextcloud to v28

mxwmnn:
didn't tested it yet, if no one tests it until the weekend i'll test :)

### DIFF
--- a/templates/requirements.yml
+++ b/templates/requirements.yml
@@ -177,7 +177,7 @@
   name: netbox
   activation_prefix: netbox_
 - src: git+https://github.com/mother-of-all-self-hosting/ansible-role-nextcloud.git
-  version: v28.0.0
+  version: v28.0.0-0
   name: nextcloud
   activation_prefix: nextcloud_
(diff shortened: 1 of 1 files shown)


spantaleev:
`requirements.yml` points to a tag of the [ansible-role-nextcloud](https://github.com/mother-of-all-self-hosting/ansible-role-nextcloud) role, not to a Nextcloud version.

Bumping the version requires:

- editing the `defaults/main.yml` file for a specific role to bump the version there (`nextcloud_version: ..`)
- commiting this
- tagging it as `VERSION-RELEASE_NUMBER` (e.g. `v28.0.0-0`)
- pushing it
- then editing `templates/requirements.yml` in this mash-playbook repository to point to the new tag for the role

I have now done all this and have adjusted the incorrect `version` (tag) in `template/requirements.yml`, so it can work.

spantaleev:
Oh, I have also upgraded my Nextcloud server from 27.1.4 to 28.0.0 and it seems to be working fine.

moan0s:
Nice. I heard some chatter that things might not work so I was hesitant to upgrade but then I'll also give the new version a try 👍

PR #135, opened 2024-01-16
Loki is ready

sudo-Tiz:
Add loki to mash.
Before merging, you have to create the 2.8.7 release for ansible-role-loki
then the loki branch can be merged to main

### DIFF
--- a/templates/group_vars_mash_servers
+++ b/templates/group_vars_mash_servers
@@ -347,6 +347,11 @@ mash_playbook_devture_systemd_service_manager_services_list_auto_itemized:
     {{ ({'name': (lago_identifier + '-pdf.service'), 'priority': 1900, 'groups': ['mash', 'lago', 'lago-pdf']} if lago_enabled else omit) }}
   # /role-specific:lago
 
+  # role-specific:loki
+  - |-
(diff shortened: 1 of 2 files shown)


moan0s:
Thanks so much for the work! When the following issues are fixed please make a PR to main (or make it right now as draft)

Here is my feedback:

- [the documentation for the service](https://github.com/sudo-Tiz/mash-playbook/blob/loki/docs/services/loki.md) is incomplete, it'd be great to have a usage section with a quick example on how to get started
- loki needs to be added to [docs/supported-services.md](https://github.com/sudo-Tiz/mash-playbook/blob/loki/docs/supported-services.md)
- the service start successfully but when trying to access loki.example.org I only get an 404 - does that work for you?
- the version numbers of our roles are `2.8.7-0` where the `-0` indicates the role iteration. It is incremented when we change something in the role without the version number of the service changing. I fixed it on this branch already

sudo-Tiz:
Thank you for the feedback !

About documentation, it is "fixed" in #138 
About the 404, there is no HTTP endpoint at / but you can go to /metrics to verify it is working

PR #141, opened 2024-01-22
change fail2ban repo

sudo-Tiz:
[etke.cc/roles/fail2ban](https://gitlab.com/etke.cc/roles/fail2ban) used by mash doesn't define the backend for sshd jail.
By default backend is set to auto and this mode doesn't work with systemd-journald.
This repository is as simple as the previous one but simply add system_security_fail2ban_sshd_backend variable to configure sshd backend and set it to systemd by default.

### DIFF
--- a/templates/requirements.yml
+++ b/templates/requirements.yml
@@ -72,8 +72,8 @@
   version: v4.97-r0-0-1
   name: exim_relay
   activation_prefix: exim_relay_
-- src: git+https://gitlab.com/etke.cc/roles/fail2ban.git
-  version: 09886730e8d3c061f22d1da4a542899063f97f0a
+- src: git+https://github.com/sudo-Tiz/ansible-role-fail2ban.git
+  version: 9f38b649c924c5aaef5785204cb93da758094a2c
   name: fail2ban
   activation_prefix: system_security_fail2ban_
 - src: git+https://github.com/mother-of-all-self-hosting/ansible-role-firezone.git


spantaleev:
If you think the `fail2ban` role is deficient, you'd better send a PR or report an Issue to it.. Instead of outright changing it with your own.

sudo-Tiz:
I agree that this would be the best solution!
But the role is on gitlab and signin-up requires a phone number and a means of payment ??!
If you have a gitlab account I strongly encourage you to propose my modification.
As it stands, the fail2ban role used by the mash-playbook is useless. 

Wouldn't the best thing for mash be to have its own roles ? Especially when it's a role that's not going to change much?

spantaleev:
You should be able to [sign up for Gitlab](https://gitlab.com/users/sign_up) for free without a phone number. You're not trying to buy their support packages or enterprise license, but merely sign up for their public free instance.

Most of the roles that make up MASH are roles developed by me or by @etkecc. Most of these roles started their life long before MASH existed, so some still live in other places (not the [mother-of-all-self-hosting](https://github.com/mother-of-all-self-hosting) Github organization). Over time, we're migrating them, so that they're all in the same place, but it's not a high priority effort.

-----

I wouldn't think the current role is useless. Maybe it's useless on certain distros which log sshd login failure attempts to `systemd` (`systemd-journald`), but a lot of distros don't do this. I'd also think that the `backend` would be auto-detected or default appropriately on the different systems. Maybe sometimes it doesn't.

It should be investigated in more detail and made to work reliably on various distros. The `fail2ban` role was developed targetting Debian-based distributions. Nevertheless, it may have gotten broken over time and we should investigate.

sudo-Tiz:
New account need phone number and payment method. Just [Searx](https://searx.space/) it.

At least, Arch Linux, which is one of the [mash recommended linux distribution](https://github.com/mother-of-all-self-hosting/mash-playbook/blob/main/docs/prerequisites.md), use journald to log sshd by default.
I'm not sure but i believe Ubuntu server minimized too. 

Anyway, i guess it could not hurt to add this option now, and then add a more automated system later.
If @etkecc see this, maybe he could update his Fail2ban role ?
Otherwise, maybe the [mother-of-all-self-hosting](https://github.com/mother-of-all-self-hosting) Github organization could migrate the role which shouldn't take more than a few minutes ☺️

PR #163, opened 2024-03-15
Improve Adguard Home's Troubleshooting section

sergiodj:
Fix grammar mistakes, improve readability and suggest "systemctl edit".

### DIFF
--- a/docs/services/adguard-home.md
+++ b/docs/services/adguard-home.md
@@ -87,16 +87,16 @@ Things you should consider doing later:
 
 ## Troubleshooting and workaround
 
-Adguard Home does currently support being setup with a non-root account (see [issue](https://github.com/AdguardTeam/AdGuardHome/issues/4714)). As the playbook uses the user `mash`, meaning you will encounter the following error when it tries to start for the first time: 
+Adguard Home does not currently support being setup with a non-`root` account (see [issue](https://github.com/AdguardTeam/AdGuardHome/issues/4714)). As the playbook uses the user `mash` when starting services, you will likely encounter the following error when `adguard-home.service` tries to start for the first time: 
 
 ```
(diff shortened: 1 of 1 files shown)


docs/services/adguard-home.md
```diff
@@ -87,16 +87,16 @@ Things you should consider doing later:
...
 ```
 
-You can get around this issue by using the following workaround which changes the user to root for the first time setup and then changes it back. You will need root to run the commands so unless you're already root, prepend the commands with `sudo` or change to root with `su`.   
+You can workaround this issue by editing `mash-adguard-home.service` and temporarily make it start Adguard Home as the `root` user for the first time, and then revert it back to using a regular user afterwards.  Follow the steps below, which require you to be `root` to execute the commands:
 
-1. Edit `/etc/systemd/system/mash-adguard-home.service` and remove or comment out the line starting with `--user` (e.g. `--user=996:3992 \` - the numbers represent the uid/gid of the `mash` user, so your values may be different):
+1. Run `systemctl edit --full mash-adguard-home.service` to edit Adguard Home's service file and remove or comment out the line starting with `--user` (e.g. `--user=996:3992 \` - the numbers represent the uid/gid of the `mash` user, so your values may be different):
```
spantaleev, line 99: Looks like you've gotten rid of the `systemctl daemon-reload` step here. I believe that one is still necessary, even when editing via `systemctl edit --full mash-adguard-home.service`

sergiodj: According to https://www.freedesktop.org/software/systemd/man/latest/systemctl.html#edit%20UNIT%E2%80%A6:

    After the units have been edited, systemd configuration is reloaded (in a way that is equivalent to daemon-reload).

I remember not needing to explicitly run `systemctl daemon-reload`, but please let me know otherwise.

spantaleev: I see! Looks like I was wrong! Looks good then!

PR #169, opened 2024-03-24
Add Tandoor service

IUCCA:
This is the same PR as before, but I accidentally deleted the old one when I created a new branch


docs/services/tandoor.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Configuration
+
+To enable this service, add the following configuration to your `vars.yml` file and re-run the [installation](../installing.md) process:
+
+```yaml
+########################################################################
+#                                                                      #
+# tandoor                                                            #
+#                                                                      #
```
spantaleev, line 21:
```suggestion
#                                                                      #
# tandoor                                                              #
#                                                                      #
```

Fix alignment

docs/services/tandoor.md
```diff
@@ -0,0 +1,54 @@
...
+tandoor_hostname: tandoor.example.com
+#path prefix is not supported at the moment
+#tandoor_path_prefix: /tandoor
+
+
+
+########################################################################
+#                                                                      #
+# /tandoor                                                           #
+#                                                                      #
```
spantaleev, line 35:
```suggestion
#                                                                      #
# /tandoor                                                             #
#                                                                      #
```

Fix alignment

docs/services/tandoor.md
```diff
@@ -0,0 +1,54 @@
...
+
+tandoor_enabled: true
+
+tandoor_hostname: tandoor.example.com
+#path prefix is not supported at the moment
+#tandoor_path_prefix: /tandoor
+
+
+
+########################################################################
```
spantaleev, line 32:
```suggestion
tandoor_hostname: tandoor.example.com

# path prefix is not supported at the moment
#tandoor_path_prefix: /tandoor

########################################################################
```

templates/group_vars_mash_servers
```diff
@@ -4458,6 +4486,77 @@ syncthing_container_labels_traefik_tls_certResolver: "{{ devture_traefik_certRes
 # /role-specific:syncthing
 
 
+
+
+########################################################################
+#                                                                      #
+# tandoor                                                            #
+#                                                                      #
```
spantaleev, line 4494:
```suggestion
#                                                                      #
# tandoor                                                              #
#                                                                      #
```

Fix alignment

templates/group_vars_mash_servers
```diff
@@ -4458,6 +4486,77 @@ syncthing_container_labels_traefik_tls_certResolver: "{{ devture_traefik_certRes
...
+tandoor_config_parameter_mail_smtpsecure: ''
+tandoor_config_parameter_mail_smtpauth: false
+tandoor_config_parameter_mail_from_address: "{{ (exim_relay_sender_address | split('@'))[0] if exim_relay_enabled else '' }}"
+tandoor_config_parameter_mail_domain: "{{ (exim_relay_sender_address | split('@'))[1] if exim_relay_enabled else '' }}"
+# /role-specific:exim_relay
+
+########################################################################
+#                                                                      #
+# /tandoor                                                           #
+#                                                                      #
```
spantaleev, line 4554:
```suggestion
#                                                                      #
# /tandoor                                                             #
#                                                                      #
```

Fix alignment

spantaleev (commented):
And some additional notes:

- indentation in [tandoor-frontend.service.j2](https://github.com/IUCCA/ansible-role-tandoor/blob/f7fcfb9ece372c2268f162894db775a8fdf53396/templates/systemd/tandoor-frontend.service.j2) seems off (mixing tabs and spaces)
- some volume mounts in [tandoor-api.service.j2](https://github.com/IUCCA/ansible-role-tandoor/blob/f7fcfb9ece372c2268f162894db775a8fdf53396/templates/systemd/tandoor-api.service.j2) can probably be `ro` (read-only)
- [uninstall.yml](https://github.com/IUCCA/ansible-role-tandoor/blob/f7fcfb9ece372c2268f162894db775a8fdf53396/tasks/uninstall.yml) does not seem to uninstall what it should (`{{ tandoor_api_identifier }}.service` and `{{ tandoor_frontend_identifier }}.service`)
- for Hubsite integration, you may also need to inject your definitions (`hubsite_service_tandoor_*`) into `mash_playbook_hubsite_service_list_auto_itemized` in `templates/group_vars_mash_servers`. Here's [an example](https://github.com/mother-of-all-self-hosting/mash-playbook/blob/7ad67129db7dc0fea5d560de2a0bb62e955cb373/templates/group_vars_mash_servers#L2456-L2467) for `adguard_home`

IUCCA:
The only volume mount in tandoor-api.service.j2 that could be ro is the nginx config. This is not ro, because tandoor has the function to change the nginx config if something changes there. Tandoor can also generate this if it is not available. However, this does not work with bind mounts, but only with docker volumes. However, since the playbook only uses bind mounts, I decided to use the nginx config with a template and still use a bind mount.

The [tandoor logo](https://github.com/TandoorRecipes/recipes/tree/5a1a5f3c4d1b696f1966ff48ad6a2e47d5045967/cookbook/static/assets) for hubsite can be found here, but there is only an svg file for the black/white one.

IUCCA:
the nginx docker container is rebuilt weekly, but uses the same tags if the nginx version has not changed. Should I therefore set tandoor_frontend_container_image_force_pull to true so that the latest updates are always installed when the playbook is executed?

spantaleev:
I think we should consider all non`-latest` and as pinned and not force-pull them. At least that's what we do for all other components.

Some of these are indeed pinned and never get rebuilt. Others may be getting rebuilt periodically. It's possibly not very important to obtain these updates. For maintained software, new versions will be issued once in a while and we'll just pull those versions whenever that happens. If Tandoor (or some other service) is not frequently issuing new versions, but is rebuilding their images on top of "newer and newer nginx" (etc.), then it may make sense to get security fixes by pulling more frequently. I suppose such an "unmaintained, yet maintained" situation is rare though. Force-pulling every time Ansible runs may be excessive.

It seems like `tandoor_frontend_version` is currently `1.25.4-alpine` and is using the official `nginx` image - not a custom one by Tandoor developers. I suppose we can just upgrade this once in a while as new nginx versions become available.

Since this webserver is only used internally (hidden behind Traefik), it's probably not that critical if it gets slightly out of date.

I think this PR looks good now, so I'm merging it!

PR #170, opened 2024-03-24
integrate just roles as pre-task on localhost

IUCCA:
This PR allows the playbook to install the roles itself via ansible galaxy, so that ansible semaphore should be able to update and run the playbook automatically.

The role also updates the templates, even if a change to the setup.yml only takes effect when the playbook is executed a second time

I have not yet tested the interaction with ansible semaphore, locally this works with ansible-playbook -i inventory/hosts setup.yml --tags=install-all,roles,start

I'm not sure if this PR is in your interest, if not, I can also spin it off into a small playbook of my own.

related to https://github.com/mother-of-all-self-hosting/ansible-role-semaphore/pull/2


spantaleev:
I see a few problems with this:

- you're relying on `roles/mash/just_roles/tasks/main.yml` to create `setup.yml` when you run `setup.yml` with `ansible-playbook`, but.. the first time around `setup.yml` doesn't exist yet
- `roles/mash/just_roles/tasks/main.yml` initializes `requirements.yml`, `setup.yml` and `group_vars/mash_servers` from the template files, which is OK, but is not ideal as it ignores the optimization feature that one could use. 

The optimization feature (see `just optimize`) tries to determine which roles you're using and then generates files (`requirements.yml`, `setup.yml` and `group_vars/mash_servers`) based on your usage. Benefits are that you have fewer roles to pull, a smaller `setup.yml` playbook and you're running fewer untrusted roles (some MASH roles are authored by 3rd parties and currently live in external repositories)

IUCCA:
I didn't know the just optimize command and the other just commands before. Would a second file in the playbook root directory next to the setup.yml be a solution that can call the individual just commands? So that the file always exists and can create and optimize the setup.yml?

Or should I simply close the PR and write my own little playbook for exactly this purpose?

spantaleev:
I suppose a new `init.yml` playbook file could run:

- `just optimize`
- `just roles`
- `just _setup-yml` 
- `just _group-vars-mash-servers`

This can also be invoked with a single `just` command: `just optimize roles _setup-yml _group-vars-mash-servers`

Doing that, you won't need to manually copy template files, etc, since these `just` recipes do it all for you.

`just optimize` is experimental, so not everyone may wish to use it yet though, so it could possibly be left out.

---

I suppose Semaphore can only be configured to invoke a single playbook, so if it invokes `init.yml` as an entrypoint, then it's `init.yml` that needs to spawn the other playbook (`setup.yml`)?

IUCCA:
I have created a first draft for an init.yml that contains the individual just commands. 

However, I see two problems when running just install-all or setup-all:

If I interpret the behavior of Semaphore correctly, the ansible-playbook is called with additional variables for authentication and is not set up globally on the host. Accordingly, you would have to extend the install-all and setup-all task with both via variables

In addition, you cannot use the web interface for the variables. Accordingly, you would have to set and manage the variables at file level.


In my opinion, a cleaner solution would therefore be to use two playbooks or two tasks in one playbook, from which the init.yml playbook initializes the other playbook and then starts the setup.yml via a semaphore API call. However, the setup.yml is called directly by ansible with the appropriate tags.

IUCCA:
The just commands do not work in the semaphore container, regardless of whether this is executed in a /tmp or a bind mount.

"just roles

error: Recipe `_ensure_file_prepared` with shebang `#!/usr/bin/env sh` execution error: Permission denied (os error 13)

error: Recipe `_requirements-yml` failed on line 121 with exit code 1"

I can't find a solution for this at the moment.


If you copy the requirements file into the roles folder and add rolename roles_path = roles/galaxy/ to ansible.cfg, semaphore can install and execute the roles. 


If the init.yml creates the files accordingly, the playbook could update itself.

Issue #171, opened 2024-03-24
gitea_container_additional_networks_custom variable is not used/working


Bergruebe:
Hi,
I'm trying to use the mash-exim-relay in Gitea. After realizing that they are not in the same network, I tried to add Gitea to the `mash-exim-relay`-network with the `gitea_container_additional_networks_custom` variable. => this is not working.

Somehow the custom variable is not used in the run of the playbook. I tested it by adding 
```yaml
- name: Debug custom network configuration
  debug:
    msg: "Error: custom: {{ gitea_container_additional_networks_custom }}  all: {{ gitea_container_additional_networks }}"
```
at the end of [`roles/gitea/tasks/install.yml`](https://github.com/mother-of-all-self-hosting/ansible-role-gitea/blob/main/tasks/install.yml). 

The custom networks is not added to  `gitea_container_additional_networks`, like it is configured [here (`ansible-role-gitea/defaults/main.yml`)](https://github.com/mother-of-all-self-hosting/ansible-role-gitea/blob/ee522640f49d310ea4227f0e9f835e5029d9b3a9/defaults/main.yml#L118-L124 ):

```yaml
# A list of additional container networks that the container would be connected to.
# The playbook does not create these networks, so make sure they already exist.
gitea_container_additional_networks: "{{ gitea_container_additional_networks_auto + gitea_container_additional_networks_custom }}"
# gitea_container_additional_networks_auto is reserved for usage by the playbook.
# Your custom networks should go into gitea_container_additional_networks_custom.
gitea_container_additional_networks_auto: []
gitea_container_additional_networks_custom: []
```

Sadly I couldn't get further with my debugging, because I don't understand how you are managing the variables in this playbook. (*Maybe a developer documentation would help 😀*)

Btw: I also tried it the other way round to add exim-relay to the gitea network, which also didn't work, see [this PR](https://github.com/mother-of-all-self-hosting/ansible-role-exim-relay/pull/1).

spantaleev:
I've merged that PR. It's good to have `exim_relay_container_additional_networks` actually working as expected.

As for integrating Gitea with exim-mailer: for other services, we usually connect the service to exim's network, not the other way around. You can see how it's done for GoToSocial here: https://github.com/mother-of-all-self-hosting/mash-playbook/blob/7ad67129db7dc0fea5d560de2a0bb62e955cb373/templates/group_vars_mash_servers#L2025-L2032

[...]

Issue #164, opened 2024-03-17
Gitea cannot complete installation (empty user)


valmar:
For reason explained in the README of the `defaults/main.yml` file, the `gitea_config_run_user` entry is left blank. However, when going to the installed gitea server, I am presented with an installation page, and I cannot proceed because the "Run as username" entry is empty, and also cannot be edited (see attached screenshot)

Thank you and keep up the good work!

![Screenshot from 2024-03-17 16-14-13](https://github.com/mother-of-all-self-hosting/mash-playbook/assets/558469/708451f3-5b01-4084-a2fe-c4fe9c3a8eab)

Bergruebe:
I had the same problem. My solution was to manually add `git` as a user in the Initial Configuration.

If I remember correctly, the field was not enabled to edit, so I needed to remove the readonly html flag via the web inspector of my browser. I know not the perfect way, but it worked somehow.

Currently I'm trying to debug why Gitea cannot send mails via mash-exim-relay. It looks like Gitea is not in the same network with the exim-relay and the variable `gitea_container_additional_networks_custom` is not working to add Gitea to the `mash-exim-relay`-network.

spantaleev:
I haven't done a new Gitea installation to confirm this yet. I suppose something may have changed on the Gitea side with regard to `gitea_config_run_user`.

-----

Or it may be that `gitea_uid` (derived from `mash_playbook_uid` via group vars) coincides with some `/etc/passwd` user inside the Gitea container.

Try `id -u mash` on the server and see what it reports. On 2 of my Gitea servers, `gitea_uid` happens to be `984` and `998`, which doesn't conflict with any existing user in the container's `/etc/passwd` file.

You can also grep for this uid in the output of `docker exec -it mash-gitea /bin/sh -c 'cat /etc/passwd'` and see if you have a match. The output currently looks like this:

```
root:x:0:0:root:/root:/bin/ash
bin:x:1:1:bin:/bin:/sbin/nologin
daemon:x:2:2:daemon:/sbin:/sbin/nologin
adm:x:3:4:adm:/var/adm:/sbin/nologin
lp:x:4:7:lp:/var/spool/lpd:/sbin/nologin
sync:x:5:0:sync:/sbin:/bin/sync
shutdown:x:6:0:shutdown:/sbin:/sbin/shutdown
halt:x:7:0:halt:/sbin:/sbin/halt
mail:x:8:12:mail:/var/mail:/sbin/nologin
[...]

PR #172, opened 2024-03-24
support exim-relay in Gitea 

Bergruebe:
Also supporting `gitea_container_additional_networks_custom`.

Solves https://github.com/mother-of-all-self-hosting/mash-playbook/issues/171

### DIFF
--- a/templates/group_vars_mash_servers
+++ b/templates/group_vars_mash_servers
@@ -1975,6 +1975,8 @@ gitea_container_additional_networks_auto: |
     ([mash_playbook_reverse_proxyable_services_additional_network] if mash_playbook_reverse_proxyable_services_additional_network else [])
     +
     ([devture_postgres_container_network] if devture_postgres_enabled and gitea_config_database_hostname == devture_postgres_identifier and gitea_container_network != devture_postgres_container_network else [])
+    +
+    ([exim_relay_container_network | default('mash-exim-relay')] if (exim_relay_enabled | default(false) and gitea_config_mailer_smtp_addr == exim_relay_identifier | default('mash-exim-relay') and gitea_container_network != exim_relay_container_network) else [])
   }}
 
 gitea_container_labels_traefik_enabled: "{{ mash_playbook_traefik_labels_enabled }}"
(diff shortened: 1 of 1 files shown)


spantaleev:
The `gitea_container_additional_networks` -> `gitea_container_additional_networks_auto` change does make sense!

I suppose you'd be adding more to this PR, so I'll wait for your call.

Bergruebe:
Yes, I'm working on it. First I need to change the Gitea role.

Bergruebe:
Do you have any idea on this issue https://github.com/mother-of-all-self-hosting/mash-playbook/issues/164

It is related to [these lines](https://github.com/mother-of-all-self-hosting/ansible-role-gitea/blob/ee522640f49d310ea4227f0e9f835e5029d9b3a9/defaults/main.yml#L44-L57):
```yaml
# Controls the RUN_USER configuration setting (`GITEA____RUN_USER` environment variable).
#
# This defaults to `git` in the upstream configuration.
#
# On startup, Gitea checks if the resolved username for the user it works with matches the RUN_USER.
#
# Because we run the container with a custom user (gitea_uid),
# and not with uid=1000 matching the git user and inside the container,
# unless we unset this, we'd get an error message:
# > `Expect user 'git' but current user is: `
# The relevant check is here: https://github.com/go-gitea/gitea/blob/ccd3a55bf4f07a390e13d48100c58fb937fd3dcf/modules/setting/setting.go#L297-L300C4
#
# Unsetting this to an empty string makes it match. Unless you gitea_uid accidentally hits some other user that is part of the container's /etc/passwd file.
gitea_config_run_user: "{{ 'git' if gitea_uid | string == '1000' else '' }}"
```

Bergruebe:
I just tested it, and it works! 

If you want to test it, you need to add`gitea_config_run_user: 'git'` to your `vars.yml`, because https://github.com/mother-of-all-self-hosting/mash-playbook/issues/164.

spantaleev:
Automatic exim-relay support seems to be working well! Thank you for these PRs!

I'll reply about #164 over there.

PR #185, opened 2024-04-08
Add Support for Wetty

spatterIight:
This PR adds support for [Wetty](https://github.com/butlerx/wetty/tree/main)

This is my first major PR for this repository, so it may be good to through every thing with a fine-toothed comb in-case I missed something by accident !

Everything seems to be working on my end, including serving on custom paths

### DIFF
--- a/docs/services/wetty.md
+++ b/docs/services/wetty.md
@@ -1,0 +1,46 @@
+# Wetty
+
+[Wetty](https://github.com/butlerx/wetty/tree/main) is an SSH terminal over HTTP/HTTPS, useful for when on a strict network which disallows outbound SSH traffic, or when only a browser can be used (like a managed chromebook).
+
+## Dependencies
+
+This service requires the following other services:
+
+- a [Traefik](traefik.md) reverse-proxy server
+
+## Configuration
+
+To enable this service, add the following configuration to your `vars.yml` file and re-run the [installation](../installing.md) process:
+
+```yaml
+########################################################################
+#                                                                      #
+# wetty                                                                #
+#                                                                      #
+########################################################################
+
+wetty_enabled: true
+wetty_hostname: mash.example.com
+wetty_path_prefix: /wetty
+wetty_ssh_host: example.com
+wetty_ssh_port: 22
+
+########################################################################
+#                                                                      #
+# /wetty                                                               #
+#                                                                      #
+########################################################################
+```
+
+In the example configuration above, we configure the service to be hosted at `https://mash.example.com/wetty` and connect to `example.com` on port `22`.
+
+You can remove the `wetty_path_prefix` variable definition, to make it default to `/`, so that the service is served at `https://mash.example.com/`.
+
+## Usage
+
+After installation, you should be able to access your new Wetty instance at: `https://WETTY_DOMAIN/PATH_PREFIX`, where:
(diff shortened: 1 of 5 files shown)


templates/group_vars_mash_servers
```diff
@@ -1481,6 +1486,43 @@ changedetection_container_labels_traefik_tls_certResolver: "{{ devture_traefik_c
...
+wetty_uid: "{{ mash_playbook_uid }}"
+wetty_gid: "{{ mash_playbook_gid }}"
+
+wetty_container_additional_networks_auto: |
+  {{
+    ([mash_playbook_reverse_proxyable_services_additional_network] if mash_playbook_reverse_proxyable_services_additional_network else [])
+  }}
+
+# role-specific:traefik
+wetty_container_labels_traefik_enabled: "{{ mash_playbook_traefik_labels_enabled }}"
```
spantaleev, line 1511: Using `role-specific` is generally a good idea, but here it might cause problems.

If the `traefik` role is not in use, then all these lines may be removed and `wetty_container_labels_traefik_enabled` will not be set in terms of `mash_playbook_traefik_labels_enabled`.

The role default for `wetty_container_labels_traefik_enabled` is also `true`: https://github.com/spatterIight/ansible-role-wetty/blob/773cd1b53f5a07e974a9f12ae64c1a5834879df3/defaults/main.yml#L48

.. so labels will be added, even if the user doesn't want them.

----

I suppose that the variables which depend on `devture_traefik_` variables may be put in a `role-specific` block, but all others (`wetty_container_labels_traefik_enabled` and `wetty_container_labels_traefik_docker_network`) should probably remain outside of it.

docs/services/wetty.md
```diff
@@ -0,0 +1,46 @@
...
+#                                                                      #
+# wetty                                                                #
+#                                                                      #
+########################################################################
+
+wetty_enabled: true
+wetty_hostname: mash.example.com
+wetty_path_prefix: /wetty
+wetty_ssh_host: example.com
+wetty_ssh_port: 22
```
spantaleev, line 26: I suppose this only works for remote machines? If one wants to connect to the local machine, how can it be done? I suppose `localhost` at port `22` won't work, since that's likely inaccessible from the wetty container.

spatterIight: You are correct that localhost will not work since that would be localhost inside the container

To connect to the same machine which is running Wetty I specify its external IP, or its domain which routes to its external IP. 

If SSH is not bound to the external interface a VPN IP (from a tunnel interface) can be used.

templates/group_vars_mash_servers
spatterIight: Went ahead and move those two variables (`wetty_container_labels_traefik_enabled` and `wetty_container_labels_traefik_docker_network`) outside of the `role-specific` block

However, I am confused since many other roles do not follow this pattern

image

Additionally, I should change the role default of `wetty_container_labels_traefik_enabled`  to `false`, correct?

spantaleev: Good point! I have adjusted all the others to follow the same pattern in e3e2aa4fb8b0d24bbfc41d5c22dbbd08811a6326

I think you can leave the role default to `true`. All other roles do the same.

PR #199, opened 2024-05-02
Add Calibre-Web role

lingawakad:
submitting a role for Calibre-Web, based off [Linuxserver.io image](https://hub.docker.com/r/linuxserver/calibre-web) - which seems to be the ["official" image](https://github.com/janeczku/calibre-web?tab=readme-ov-file#docker-images).

unfortunately, unable to run rootless, but is set to run as the mash user internally like the LSIO images are able to.

### DIFF
--- a/docs/services/calibre_web.md
+++ b/docs/services/calibre_web.md
@@ -18,14 +18,14 @@ To enable this service, add the following configuration to your `vars.yml` file 
 ```yaml
 ########################################################################
 #                                                                      #
-# calibre_web                                                            #
+# calibre-web                                                            #
 #                                                                      #
 ########################################################################
 
 calibre_web_enabled: true
 
 calibre_web_hostname: mash.example.com
-calibre_web_path_prefix: /calibre_web
+calibre_web_path_prefix: /calibre-web
 
 # By default, calibre_web will look at the /books directory for your Calibre database.
 #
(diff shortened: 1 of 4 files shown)


moan0s:
Thank you a lot for the PR! As mentioned in the issue I tested the role and it looks good to me. One detail on the versioning: you currently tagged the role with the calibre-web version `v0.6.21`. We usually append a `-<number>` at the back so `v0.6.21-0` so that we can have multiple tags on the role per version. E.g. we add an additional enviroment variable but calibre-web does not have a new version: just increase "our" number to `v0.6.21-1`.
This will help the `just roles` and `just update` to find the correct version. Other than that, maybe @spantaleev wants to have a look but from my perspective this is ready to be merged

docs/services/calibre_web.md
```diff
@@ -0,0 +1,136 @@
...
+
+The default username is `admin` and the default password is `admin123`.
+You'll be able to change the username and password, and add additional users in the web UI.
+
+On the initial setup screen, enter /books as your calibre library location.
+If you haven't placed a Calibre database in that directory on the host yet, it will error as an invalid location.
+
+### Syncthing integration
+
+If you've got a [Syncthing](syncthing.md) service running, you can use it to synchronize your music directory onto the server and then mount it as read-only into the calibre_web container.
```
spantaleev, line 62: "music directory"

lingawakad: ah, missed one, thanks

PR #204, opened 2024-05-15
Paperless OCR Languages

JokerGermany:
If the documents aren't in english it's necessary to add these commands.
I think because this is an important part of paperless this should be in the installation guide.

e.g:
If you use the default settings german umlauts aren't recognized.

If you don't like the wording, feel free to change the text =)

### DIFF
--- a/docs/services/paperless-ngx.md
+++ b/docs/services/paperless-ngx.md
@@ -178,6 +178,24 @@ paperless_container_additional_networks_custom:
 #                                                                      #
 ########################################################################
 ```
+### Languages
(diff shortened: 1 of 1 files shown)


JokerGermany:
I assume that this command 
```
paperless_environment_variables_extension: |
   PAPERLESS_OCR_LANGUAGES=tur ces chi-tra
```
only work, when this PR is merged:
https://github.com/mother-of-all-self-hosting/ansible-role-paperless/pull/5

I assume it don't work with --read-only

spantaleev:
Looks like this has been done (in a slightly different way) via ca54fad95b7f5d57533349c2ff85b7c3eb2c6dec

PR #205, opened 2024-05-18
Add developer documentation

Bergruebe:
I tried to create an ansible-role for wallabag, and it was possible for me, who is still quite new to ansible, by just looking into the existing files and roles you already created. But it would have been much easier if there was a documentation on which steps are required to create and add a new service to this great playbook.

I'm not sure if any steps are required for `bin/versions.py` and `bin/optimize.py`.

It would be really great if someone could document how to use traefik in a role. I don't have the knowledge to do that.

### DIFF
--- a/docs/developer-documentation.md
+++ b/docs/developer-documentation.md
@@ -1,0 +1,216 @@
+# Developer Documentation
+
+## Support a new service | Create your own role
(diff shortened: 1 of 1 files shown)


spantaleev:
Thanks!

PR #216, opened 2024-06-20
Bump Calibre-Web role to v0.6.21-1

lingawakad:
Adds support for LSIO Docker mods


moan0s:
This version could provoke this bug: https://github.com/lingawakad/ansible-role-calibre-web/issues/2 but could also be unrelated. Maybe a bit more testing/feedback is needed before merging

mwelsch:
I just tested it and had no issue. Also tested to change the password with the python script and worked fine for me

PR #218, opened 2024-06-21
Adds Matterbridge service

oliverlorenz:
I have implemented Matterbridge as a service and would like to contribute it.

As this is my first public MASH role, I would be very happy to have some feedback. If it is then considered okay, I would prefer the role to move into the MASH organisation.

I have also integrated the role into the mash-playbook. But I'm not sure if I've thought of everything. I would also be grateful for feedback on this :) 

Best regards
Oliver

### DIFF
--- a/docs/services/matterbridge.md
+++ b/docs/services/matterbridge.md
@@ -1,0 +1,23 @@
+# Matterbridge
+
+[Matterbridge](https://github.com/42wim/matterbridge)
+
+## Configuration
+
+To enable this service, add the following configuration to your `vars.yml` file and re-run the [installation](../installing.md) process:
+
+```yaml
+########################################################################
+#                                                                      #
+# matterbridge                                                         #
+#                                                                      #
+########################################################################
+
+matterbridge_enabled: true
+matterbridge_toml_src_path: "{{ playbook_dir }}/path/to/your/matterbridge.toml.j2"
+########################################################################
+#                                                                      #
+# /matterbridge                                                        #
+#                                                                      #
+########################################################################
+```
--- a/docs/services/matterbridge.md
+++ b/docs/services/matterbridge.md
@@ -6,6 +6,8 @@
 
 To enable this service, add the following configuration to your `vars.yml` file and re-run the [installation](../installing.md) process:
 
+The configuration itself is documented [here](https://github.com/42wim/matterbridge/wiki/How-to-create-your-config)
+
 ```yaml
 ########################################################################
 #                                                                      #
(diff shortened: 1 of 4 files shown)


docs/services/matterbridge.md
```diff
@@ -0,0 +1,23 @@
...
+
+```yaml
+########################################################################
+#                                                                      #
+# matterbridge                                                         #
+#                                                                      #
+########################################################################
+
+matterbridge_enabled: true
+matterbridge_toml_src_path: "{{ playbook_dir }}/path/to/your/matterbridge.toml.j2"
```
spantaleev, line 17: It's probably better to have a `matterbridge_configuration_toml` variable.

People can populate it either inline (in `vars.yml`) or using the Ansible [lookup](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_lookups.html) module (for people who wish to store it in a separate file).

docs/services/matterbridge.md
```diff
@@ -0,0 +1,23 @@
+# Matterbridge
+
+[Matterbridge](https://github.com/42wim/matterbridge)
+
+## Configuration
+
```
spantaleev, line 6: Perhaps we should point people to https://github.com/42wim/matterbridge/wiki/How-to-create-your-config for tips on how they can generate the Matterbridge configuration.

templates/setup.yml
```diff
@@ -411,6 +411,10 @@
     - role: galaxy/auxiliary
     # /role-specific:auxiliary
 
+    # role-specific:matterbridge
+    - role: galaxy/matterbridge
```
spantaleev, line 415: Some comments on the role itself:

- in `templates/matterbridge.service.j2`, the `--mount type=bind,src={{ matterbridge_toml_path }},dst=/etc/matterbridge/matterbridge.toml \` line could probably benefit from `,readonly`. If the bridge needs the file to be writable, perhaps it can be documented there.

- if the Matterbridge container can start as `--read-only`, it'd be good to run it that way. If not, please document in `matterbridge.service` why it's not done. Sometimes mounting a tmpfs somewhere may let a service start in a readonly container ([example from Plausible](https://github.com/mother-of-all-self-hosting/ansible-role-plausible/blob/d245e96d0f88e22892daf34d495ecc28b5ee5206/templates/systemd/plausible.service.j2#L35)), while other times it's not possible.

- there are no systemd requires/wants definitions for `matterbridge.service`. We tend to make all our roles depend on `docker.service`. Recently, we started using `devture_systemd_docker_base_docker_service_name` instead of hardcoding `docker.service`. Refer to the [Plausible role](https://github.com/mother-of-all-self-hosting/ansible-role-plausible) for an example ([here](https://github.com/mother-of-all-self-hosting/ansible-role-plausible/blob/d245e96d0f88e22892daf34d495ecc28b5ee5206/defaults/main.yml#L194-L204) and [here](https://github.com/mother-of-all-self-hosting/ansible-role-plausible/blob/d245e96d0f88e22892daf34d495ecc28b5ee5206/templates/systemd/plausible.service.j2#L4-L11)).

- the Matterbridge link in `README.md` is currently broken

docs/services/matterbridge.md
oliverlorenz: Done :slightly_smiling_face:

templates/setup.yml
oliverlorenz: I have incorporated your feedback and believe I have implemented it as you intended. Would you mind taking another look at it?

docs/services/matterbridge.md
oliverlorenz: done :heavy_check_mark:

templates/setup.yml
spantaleev: It looks good now!

The only thing I've noticed is that `matterbridge_config_dir_path` is actually pointed to a directory named `data` which is confusing.

Many of the roles have an Ansible-managed config (dumped into a `config` directory), as well as some `data` directory which contains container-generated data. Naming the config directory `data` is a source of confusion.

It'd be great to change this now, before people start using it (and having to worry about a migration path).

oliverlorenz: I agree that this is confusing. I changed this in the role repository :+1:

oliverlorenz:
It seems like all the tasks have been completed. What are the next steps? I would still prefer if we could push the repository into the "mother-of-all-self-hosting" setup.

spantaleev:
Thank you! It's merged now!

I've relocated the role to the MASH organization as well.

PR #225, opened 2024-07-04
add Cadvisor

sudo-Tiz:
role is working but :
can't run rootless
path_prefix doesn't work (it's currently commented in the group_vars)

please also consider cloning [ansible-role-cadvisor](https://github.com/sudo-Tiz/ansible-role-cadvisor) and updating url in requierements.yml to let mash own the role (and use your automation to keep it updated)

### DIFF
--- a/templates/group_vars_mash_servers
+++ b/templates/group_vars_mash_servers
@@ -1610,20 +1610,10 @@ wetty_container_labels_traefik_tls_certResolver: "{{ devture_traefik_certResolve
 
 cadvisor_enabled: false
 
-cadvisor_identifier: "{{ mash_playbook_service_identifier_prefix }}prometheus-node-exporter"
-
-cadvisor_hostname: "{{ mash_playbook_metrics_exposure_hostname }}"
-cadvisor_path_prefix: "{{ mash_playbook_metrics_exposure_path_prefix }}/{{ cadvisor_identifier }}"
-
-cadvisor_base_path: "{{ mash_playbook_base_path }}/{{ mash_playbook_service_base_directory_name_prefix }}prometheus-node-exporter"
+cadvisor_identifier: "{{ mash_playbook_service_identifier_prefix }}cadvisor"
 
 cadvisor_uid: "{{ mash_playbook_uid }}"
 cadvisor_gid: "{{ mash_playbook_gid }}"
-
-cadvisor_container_additional_networks: |
-  {{
-    ([mash_playbook_reverse_proxyable_services_additional_network] if mash_playbook_reverse_proxyable_services_additional_network else [])
-  }}
 
 # Only enable Traefik labels if a hostname is set (indicating that this will be exposed publicly)
 cadvisor_container_labels_traefik_enabled: "{{ mash_playbook_traefik_labels_enabled and cadvisor_hostname }}"
@@ -1631,18 +1621,22 @@ cadvisor_container_labels_traefik_docker_network: "{{ mash_playbook_reverse_prox
 cadvisor_container_labels_traefik_entrypoints: "{{ devture_traefik_entrypoint_primary }}"
 cadvisor_container_labels_traefik_tls_certResolver: "{{ devture_traefik_certResolver_primary }}"
 
+cadvisor_container_labels_metrics_enabled: "{{ prometheus_enabled | default(false) or mash_playbook_metrics_exposure_enabled }}"
+cadvisor_container_labels_metrics_hostname: "{{ mash_playbook_metrics_exposure_hostname }}"
+cadvisor_container_labels_metrics_path_prefix: "{{ mash_playbook_metrics_exposure_path_prefix }}/{{ cadvisor_identifier }}"
+cadvisor_container_labels_metrics_traefik_middleware_basic_auth_enabled: "{{ mash_playbook_metrics_exposure_http_basic_auth_enabled }}"
+cadvisor_container_labels_metrics_traefik_middleware_basic_auth_users: "{{ mash_playbook_metrics_exposure_http_basic_auth_users }}"
 cadvisor_container_labels_metrics_middleware_basic_auth_enabled: "{{ mash_playbook_metrics_exposure_http_basic_auth_enabled }}"
 cadvisor_container_labels_metrics_middleware_basic_auth_users: "{{ mash_playbook_metrics_exposure_http_basic_auth_users }}"
 
-cadvisor_container_extra_arguments_auto:
-  - "--mount type=bind,source=/,destination=/rootfs,readonly"
-  - "--mount type=bind,source=/var/run,destination=/var/run,readonly"
-  - "--mount type=bind,source=/sys,destination=/sys,readonly"
-  - "--mount type=bind,source=/var/lib/docker/,destination=/var/lib/docker,readonly"
-  - "--mount type=bind,source=/dev/disk/,destination=/dev/disk,readonly"
-  - "--mount type=bind,source=/dev/kmsg,target=/dev/kmsg"
-
-
+cadvisor_container_additional_networks_auto: |
+  {{
+    (
+      ([mash_playbook_reverse_proxyable_services_additional_network] if mash_playbook_reverse_proxyable_services_additional_network else [])
+      +
+      ([prometheus_container_network] if (prometheus_enabled | default(false) and prometheus_container_network | default('') != cadvisor_container_network) else [])
+    ) | unique
+  }}
 
 cadvisor_process_extra_arguments_auto:
   - "--url_base_prefix={{ cadvisor_container_labels_traefik_path_prefix }}"
--- a/templates/group_vars_mash_servers
+++ b/templates/group_vars_mash_servers
@@ -186,6 +186,11 @@ mash_playbook_devture_systemd_service_manager_services_list_auto_itemized:
     {{ ({'name': (authelia_identifier + '.service'), 'priority': 2000, 'groups': ['mash', 'authelia']} if authelia_enabled else omit) }}
   # /role-specific:authelia
 
+  # role-specific:cadvisor
+  - |-
+    {{ ({'name': (cadvisor_identifier + '.service'), 'priority': 500, 'groups': ['mash', 'metrics', 'prometheus-node-exporter']} if cadvisor_enabled else omit) }}
+  # /role-specific:cadvisor
+
   # role-specific:calibre-web
   - |-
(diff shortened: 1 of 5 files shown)


spantaleev:
Some feedback on the role:

- the "Project source code URL" in `defaults/main.yml` is incorrect
- `defaults/main.yml` mentions "node_exporter" / "node exporter" in multiple places
- if the role only exposes metrics on the `cadvisor_hostname`, then there's no need for additional labels for XSS protection and whatnot (`cadvisor_container_labels_traefik_additional_response_headers_auto`). Feel free to keep this "additional response headers" feature, but you can remove the `_auto` stuff (and leave it as an empty list). It seems like `_auto` refers to variables like `cadvisor_http_header_xss_protection`, `cadvisor_http_header_frame_options`, etc., which don't exist. It seems like neither `cadvisor_container_labels_traefik_additional_request_headers`, nor `cadvisor_container_labels_traefik_additional_response_headers` are actually used in `templates/labels.j2`, so.. maybe you should actually get rid of all of these variables.
- "exposes its HTTP port (tcp/9100 in the container)" seems to be incorrect, given that `cadvisor_container_http_port: 8080`. This needs to be adjusted or the `(..)` part can be removed.

spantaleev:
The documentation page mentions `cadvisor_hostname`, but group vars are auto-setting `cadvisor_container_labels_metrics_hostname` directly. This is probably a bit odd.

For other roles, we do it like this, because the main hostname (`SERVICE_hostname`) is actually used for a web UI (and as a default for the metrics hostname).

For this role, there's no web UI and you're telling people to use `cadvisor_hostname` for configuring the metrics hostname, so.. maybe it makes sense to have group vars wiring influence `cadvisor_hostname` (which then goes to influence `cadvisor_container_labels_metrics_hostname` via `defaults/main.yml`).

sudo-Tiz:
Thank you for your feedback, @spantaleev. 

I recognized that the role was far from ready for merging as it contained copy/paste artifacts from the initial role. I have since updated the documentation and made changes to the role accordingly. 

Regarding cAdvisor, I have removed the "metrics" labels since it exposes both metrics and webUI at the same address and port

 I believe the role should now be ready for testing

templates/group_vars_mash_servers
```diff
@@ -186,6 +186,11 @@ mash_playbook_devture_systemd_service_manager_services_list_auto_itemized:
     {{ ({'name': (authelia_identifier + '.service'), 'priority': 2000, 'groups': ['mash', 'authelia']} if authelia_enabled else omit) }}
   # /role-specific:authelia
 
+  # role-specific:cadvisor
+  - |-
+    {{ ({'name': (cadvisor_identifier + '.service'), 'priority': 500, 'groups': ['mash', 'metrics', 'prometheus-node-exporter']} if cadvisor_enabled else omit) }}
```
spantaleev, line 191:
```suggestion
    {{ ({'name': (cadvisor_identifier + '.service'), 'priority': 500, 'groups': ['mash', 'metrics', 'cadvisor']} if cadvisor_enabled else omit) }}
```

templates/group_vars_mash_servers
```diff
@@ -1596,6 +1601,55 @@ wetty_container_labels_traefik_tls_certResolver: "{{ devture_traefik_certResolve
...
+########################################################################
+
+cadvisor_enabled: false
+
+cadvisor_identifier: "{{ mash_playbook_service_identifier_prefix }}cadvisor"
+
+cadvisor_uid: "{{ mash_playbook_uid }}"
+cadvisor_gid: "{{ mash_playbook_gid }}"
+
+# Only enable Traefik labels if a hostname is set (indicating that this will be exposed publicly)
```
spantaleev, line 1618:
```suggestion
cadvisor_uid: "{{ mash_playbook_uid }}"
cadvisor_gid: "{{ mash_playbook_gid }}"

cadvisor_base_path: "{{ mash_playbook_base_path }}/{{ mash_playbook_service_base_directory_name_prefix }}cadvisor"

# Only enable Traefik labels if a hostname is set (indicating that this will be exposed publicly)
```

Without this override,  cAdvisor uses the role defaults and gets installed to `/mash-cadvisor`

templates/group_vars_mash_servers
```diff
@@ -1596,6 +1601,55 @@ wetty_container_labels_traefik_tls_certResolver: "{{ devture_traefik_certResolve
...
+cadvisor_container_labels_traefik_entrypoints: "{{ devture_traefik_entrypoint_primary }}"
+cadvisor_container_labels_traefik_tls_certResolver: "{{ devture_traefik_certResolver_primary }}"
+
+cadvisor_container_labels_metrics_enabled: "{{ prometheus_enabled | default(false) or mash_playbook_metrics_exposure_enabled }}"
+cadvisor_container_labels_metrics_hostname: "{{ mash_playbook_metrics_exposure_hostname }}"
+cadvisor_container_labels_metrics_path_prefix: "{{ mash_playbook_metrics_exposure_path_prefix }}/{{ cadvisor_identifier }}"
+cadvisor_container_labels_metrics_traefik_middleware_basic_auth_enabled: "{{ mash_playbook_metrics_exposure_http_basic_auth_enabled }}"
+cadvisor_container_labels_metrics_traefik_middleware_basic_auth_users: "{{ mash_playbook_metrics_exposure_http_basic_auth_users }}"
+cadvisor_container_labels_metrics_middleware_basic_auth_enabled: "{{ mash_playbook_metrics_exposure_http_basic_auth_enabled }}"
+cadvisor_container_labels_metrics_middleware_basic_auth_users: "{{ mash_playbook_metrics_exposure_http_basic_auth_users }}"
```
spantaleev, line 1630: These variables do not seem to be defined anymore, yet.. they're here.

That said, I think it's better if  metrics had their own Traefik router (separate from the web UI) and for them to respect the `mash_playbook_metrics_exposure_*` variables automatically (auto-enabling metrics exposure for this service, possibly protected with the Basic Auth credentials specified in `mash_playbook_metrics_exposure_http_basic_auth_*`).

The web UI could remain optional and have its (optional) separate set of Basic Auth credentials

templates/group_vars_mash_servers
```diff
@@ -1596,6 +1601,55 @@ wetty_container_labels_traefik_tls_certResolver: "{{ devture_traefik_certResolve
...
+
+cadvisor_container_additional_networks_auto: |
+  {{
+    (
+      ([mash_playbook_reverse_proxyable_services_additional_network] if mash_playbook_reverse_proxyable_services_additional_network else [])
+      +
+      ([prometheus_container_network] if (prometheus_enabled | default(false) and prometheus_container_network | default('') != cadvisor_container_network) else [])
```
spantaleev, line 1637: Our other roles do not auto-connect to the Prometheus network.

We try to minimize such auto-wiring in MASH to keep things more predictable. Users are expected to hook Prometheus to everything it needs to scrape by adjusting its own "additional networks".

We only do auto-wiring with Traefik and databases (Postgres, MariaDB).

templates/group_vars_mash_servers
```diff
@@ -1596,6 +1601,55 @@ wetty_container_labels_traefik_tls_certResolver: "{{ devture_traefik_certResolve
...
+    ) | unique
+  }}
+
+cadvisor_process_extra_arguments_auto:
+  - "--url_base_prefix={{ cadvisor_container_labels_traefik_path_prefix }}"
```
spantaleev, line 1642:
```suggestion
  - "--url_base_prefix={{ cadvisor_path_prefix }}"
```

Using the top-level `cadvisor_path_prefix` variable directly is probably better.

spantaleev (changes requested):
I was thinking of pushing this over the finish line, but it has tons of issues.

Besides the ones reported in inline comments in the review, there are also these:

- the `cadvisor_config_providers_docker_endpoint_is_unix_socket` variable in `vars/main.yml` has some Traefik mentions in its comments. Not sure if it's relevant to cAdvisor or not
- the systemd service file (`templates/cadvisor.service.j2`) does `--cap-drop=ALL` only when a non-unix-socket is used. Not sure if this is intentional or a mistake
- the cAdvisor container seems stuck in an unhealthy state, which makes Traefik not want to send requests to it. The docs page seems to mention an additional environment variable (`CADVISOR_HEALTHCHECK_URL=http://localhost:8080/healthz`), but.. adding this does not help either. If this is a good value, it should be default anyway - why ask users to specify it manually? Regardless of what i did, the container was unhealthy, which may be due to a misconfiguration. 
- the docs page says "You will have to mount specific folders depending on your need", but.. it's not clear what a good default is..  I tried mounting whatever you're recommending (except for `/dev/disk/`) and cAdvisor still fails to start in a healthy state
- `cadvisor_dashboard_urls` points to some dashboard, but I don't see it being mentioned in the docs (with instructions how to hook it to Grafana). For some example instructions, see `docs/services/grafana.md`

spantaleev:
I was thinking of pushing this over the finish line, but it has tons of issues:

- the `cadvisor_config_providers_docker_endpoint_is_unix_socket` variable in `vars/main.yml` has some Traefik mentions in its comments. Not sure if it's relevant to cAdvisor or not
- the systemd service file (`templates/cadvisor.service.j2`) does `--cap-drop=ALL` only when a non-unix-socket is used. Not sure if this is intentional or a mistake
- the cAdvisor container seems stuck in an unhealthy state, which makes Traefik not want to send requests to it. The docs page seems to mention an additional environment variable (`CADVISOR_HEALTHCHECK_URL=http://localhost:8080/healthz`), but.. adding this does not help either. If this is a good value, it should be default anyway - why ask users to specify it manually? Regardless of what i did, the container was unhealthy, which may be due to a misconfiguration. 
- the docs page says "You will have to mount specific folders depending on your need", but.. it's not clear what a good default is..  I tried mounting whatever you're recommending (except for `/dev/disk/`) and cAdvisor still fails to start in a healthy state
- `cadvisor_dashboard_urls` points to some dashboard, but I don't see it being mentioned in the docs (with instructions how to hook it to Grafana). For some example instructions, see `docs/services/grafana.md`
- it would have been better

templates/group_vars_mash_servers
sudo-Tiz: Variables removed.

The Web UI is accessible at the root endpoint (/) and is available on the same port as the metrics, which can be found at /metrics. The Web UI provides the same information but integrates Google Analytics to create a more user-friendly experience. 

That said, I believe that the metrics and the Web UI do not require separate Traefik routers.

spatterIight:
This variable in the role is defined but never used i think:

```yaml
# Controls the `providers.docker.network` configuration option.
cadvisor_config_providers_docker_network: "{{ devture_traefik_container_network }}"
```

image

PR #228, opened 2024-07-06
Add Endlessh-go

sudo-Tiz:


docs/services/endlessh.md
```diff
@@ -0,0 +1,70 @@
+# Endlessh
+
+[Endlessh-go](https://github.com/shizunge/endlessh-go) is a A golang implementation of endlessh, a ssh trapit. Installing it is powered by the [mother-of-all-self-hosting/ansible-role-endlessh](https://github.com/mother-of-all-self-hosting/ansible-role-endlessh) Ansible role.
```
spantaleev, line 3: https://github.com/mother-of-all-self-hosting/ansible-role-endlessh does not exist.

Or would you like us to take your own role and put it under the MASH organization?

-----

"a SSH trapit" is a typo. Try "an SSH tarpit".

It could also link to https://nullprogram.com/blog/2019/03/22/

There are some more typos in this sentence "a A golang" -> "a Golang implementation".

"endlessh" could be a link to the original C implementation (https://github.com/skeeto/endlessh).

docs/services/endlessh.md
```diff
@@ -0,0 +1,70 @@
...
+#                                                                      #
+# /endlessh                                                                #
+#                                                                      #
+########################################################################
+```
+See the full list of options in the [default/main.yml](default/main.yml) file
+
+## Exporting Prometheus metrics
+
+Default port is 2112
```
spantaleev, line 40: Maybe we need more documentation about how scraping should be configured.

None of the MASH roles automatically wire Prometheus to scrape them, and this one is no exception.

It seems like at the bare minimium, people would need to ensure Prometheus (if the one managed by the playbook is used) is on the same container network as endlessh (either by editing `endlessh_container_additional_networks_custom` or by editing `prometheus_container_additional_networks_auto`).

Although.. if both Prometheus and endlessh are exposed via Traefik, they already shared a network (`traefik`).

docs/services/endlessh.md
```diff
@@ -0,0 +1,70 @@
...
+
+Default entrypoint is /metrics
+
+use environment variable 
+```yaml
+endlessh_container_extra_arguments_custom:
+  - "-enable_prometheus"
+```
```
spantaleev, line 48: This seems like an extra flag, not an environment variable.

docs/services/endlessh.md
```diff
@@ -0,0 +1,70 @@
...
+To enable this service, add the following configuration to your `vars.yml` file and re-run the [installation](../installing.md) process:
+
+```yaml
+########################################################################
+#                                                                      #
+# endlessh                                                                 #
+#                                                                      #
+########################################################################
+
+endlessh_enabled: true
```
spantaleev, line 28: Maybe we should mention the `endlessh_container_host_bind_port: 22` default and the fact that endlessh will try to bind to port 22 on all network interfaces by default.

This may be somewhat obvious, but it's probably good to mention it explicitly.

docs/services/endlessh.md
```diff
@@ -0,0 +1,70 @@
...
+
+########################################################################
+#                                                                      #
+# /endlessh                                                                #
```
spantaleev, line 32: Alignment issue

docs/services/endlessh.md
```diff
@@ -0,0 +1,70 @@
...
+To configure and install endlessh on your own server(s), you should use a playbook like [Mother of all self-hosting](https://github.com/mother-of-all-self-hosting/mash-playbook) or write your own.
+
+## Configuration
+
+To enable this service, add the following configuration to your `vars.yml` file and re-run the [installation](../installing.md) process:
+
+```yaml
+########################################################################
+#                                                                      #
+# endlessh                                                                 #
```
spantaleev, line 24: Alignment issue

docs/services/endlessh.md
```diff
@@ -0,0 +1,70 @@
...
+
+use environment variable 
+```yaml
+endlessh_container_extra_arguments_custom:
+  - "-enable_prometheus"
+```
+
+### Exposing the web interface
+
+By setting a hostname and optionally a path prefix, you can expose endlessh publicly.
```
spantaleev, line 52: Does not setting up a hostname disable exposure? Or will it complain that it needs it?

Looking at the code, it seems like it's completely optional. We should make this more explicit.

-----

Will the web UI show useful stuff even without Prometheus and Grafana? I suppose it would.

templates/group_vars_mash_servers
```diff
@@ -1932,6 +1937,43 @@ echoip_container_labels_traefik_tls_certResolver: "{{ devture_traefik_certResolv
...
 
+# role-specific:endlessh
+########################################################################
+#                                                                      #
+# endlessh                                                             #
+#                                                                      #
+########################################################################
+
+endlessh_enabled: false
+endlessh_hostname: "{{ mash_playbook_metrics_exposure_hostname }}"
```
spantaleev, line 1948: Hmm.. `endlessh_hostname` is used for the endlessh web UI, not just for metrics exposure, right.

This seems like an unfortunate default.

Because of it: people who don't define `endlessh_hostname` (thinking it won't expose the web UI), but use `mash_playbook_metrics_exposure_hostname` (for other reasons) will be surprised by the fact that the endlessh web UI is actually exposed on the "global metrics hostname". Not only that, but since they never intended to expose its web UI, they likely would not have configured Basic Authentication (`endlessh_container_labels_middleware_basic_auth_*`) and will have the web UI served publicly.

-----

I think that:

- `endlessh_hostname` and `endlessh_path_prefix` should not have default values

- the role should get another set of `_traefik` variables for handling metrics and their exposure (see [the group vars for promtail](https://github.com/mother-of-all-self-hosting/mash-playbook/blob/3c18123de9ab76139e4ef7539bf58dfef79cabc6/templates/group_vars_mash_servers#L4710-L4720) for an example)

This way, the defaults are sane and secure.. And metrics may be exposed on a completely different hostname and path prefix and with completely different Basic Auth credentials (auto-wired via `mash_playbook_metrics_exposure_http_basic_auth_enabled` and `mash_playbook_metrics_exposure_http_basic_auth_users`).

docs/services/endlessh.md
```diff
@@ -0,0 +1,70 @@
...
+
+[Endlessh-go](https://github.com/shizunge/endlessh-go) is a A golang implementation of endlessh, a ssh trapit. Installing it is powered by the [mother-of-all-self-hosting/ansible-role-endlessh](https://github.com/mother-of-all-self-hosting/ansible-role-endlessh) Ansible role.
+
+## Dependencies
+
+This service requires the following other services:
+
+- (optionally) [Traefik](traefik.md) - a reverse-proxy server for exposing endlessh publicly
+- (optionally) [Prometheus](./prometheus.md) - a database for storing metrics
+- (optionally) [Grafana](./grafana.md) - a web UI that can query the prometheus datasource (connection) and display the logs
```
spantaleev, line 11: I suppose a valid pre-requisite for any sort of SSH tarpit system is to:

- either: not have an SSH server at all (likely not the case for many)
- or: relocating the real SSH server to another port (freeing up the default port `22` for the tarpit server)

We should probably indicate this somewhere. Maybe in a new "Prerequisites" section, because it's not really a dependency.

templates/group_vars_mash_servers
```diff
@@ -1932,6 +1937,43 @@ echoip_container_labels_traefik_tls_certResolver: "{{ devture_traefik_certResolv
...
+
+
+# Only enable Traefik labels if a hostname is set (indicating that this will be exposed publicly)
+endlessh_container_labels_traefik_enabled: "{{ mash_playbook_traefik_labels_enabled and endlessh_hostname }}"
+endlessh_container_labels_traefik_docker_network: "{{ mash_playbook_reverse_proxyable_services_additional_network }}"
+endlessh_container_labels_traefik_entrypoints: "{{ devture_traefik_entrypoint_primary }}"
+endlessh_container_labels_traefik_tls_certResolver: "{{ devture_traefik_certResolver_primary }}"
+
+endlessh_container_labels_metrics_middleware_basic_auth_enabled: "{{ mash_playbook_metrics_exposure_http_basic_auth_enabled }}"
+endlessh_container_labels_metrics_middleware_basic_auth_users: "{{ mash_playbook_metrics_exposure_http_basic_auth_users }}"
```
spantaleev, line 1968: These variables do not seem to exist yet, but.. as mentioned in another comment in this review, separate metrics-related Traefik labels variables should be added, so as to:

- allow the other labels to only handle the web UI
- allow these new variables (and associated labels) to only deal with metrics.. and to allow metrics to be on their own hostname, path prefix and use their own Basic Auth credentials

docs/services/endlessh.md
sudo-Tiz: Endlessh-go has no Web UI, `endlessh_hostname` is only used for metrics exposure.
Therefor, setting a Hostname is only to required to expose metrics publicly.
I Fixed the documentation.

docs/services/endlessh.md
sudo-Tiz: Please take my role and put it under the MASH organization.

templates/group_vars_mash_servers
sudo-Tiz: I believe these variables exist in the role's [templates/labels.j2](https://github.com/sudo-Tiz/ansible-role-endlessh/blob/main/templates/labels.j2) and in the role's [default/main](https://github.com/sudo-Tiz/ansible-role-endlessh/blob/main/defaults/main.yml)

sudo-Tiz:
Thank you for your feedback @spantaleev 
I upgraded the documentation to take account of your comments. 
I hope the documentation is clear enough now to avoid confusion. 

Endlessh-go has no Web UI, `endlessh_hostname` is only used for metrics exposure. Therefore, There are no needs of separate metrics-related Trafik laebls variables and label variables should look like [Prometheus postgres exporter](https://github.com/mother-of-all-self-hosting/mash-playbook/blob/main/docs/services/prometheus-postgres-exporter.md) more than Promtail (which has a Web UI).

Please tell me if something is still missing/wrong in the role.

Please take my role and put it under the MASH organization.

Issue #12, opened 2023-03-20
Service wishlist: What services would be nice to support


aine-etke:
- [x] Ansible Semaphore
- [x] appsmith
- [x] audiobookshelf — https://github.com/mother-of-all-self-hosting/mash-playbook/pull/1076
- [x] Calibre Web
- [x] Cryptpad
- [x] endlessh-go
- [x] etke.cc/cleanup (requires adjustments)
- [x] etke.cc/etherpad
- [x] etke.cc/grafana
- [x] etke.cc/languagetool
- [x] etke.cc/ntfy
- [x] etke.cc/prometheus_blackbox_exporter (requires full overhaul)
- [x] etke.cc/prometheus_postgres_exporter
- [x] etke.cc/security (requires adjustments)
- [x] etke.cc/soft-serve (requires full overhaul)
- [x] etke.cc/swap (requires adjustments)
- [x] Excalidraw
- [x] filebrowser
- [x] Healthchecks.io
- [x] Homarr
- [x] Immich
- [x] jellyfin — https://github.com/mother-of-all-self-hosting/mash-playbook/pull/464
- [x] Jellyseerr — https://github.com/mother-of-all-self-hosting/mash-playbook/pull/1026
- [x] Joplin-server
- [x] Listmonk
- [x] lldap — https://github.com/mother-of-all-self-hosting/mash-playbook/pull/978
- [x] Loki
- [x] Matomo
- [x] mdad/prometheus (requires full overhaul)
- [x] Mobilizon
- [x] OSM Tiling server
- [x] Otterwiki — https://github.com/mother-of-all-self-hosting/mash-playbook/pull/1328
- [x] Paperless
- [x] Prometheus Alertmanager ([ansible-role-alertmanager](https://github.com/mother-of-all-self-hosting/ansible-role-alertmanager))
- [x] RSS-Bridge — https://github.com/mother-of-all-self-hosting/mash-playbook/pull/1143
- [x] SearxNG
- [x] Vikunja
- [x] Weblate - https://github.com/mother-of-all-self-hosting/mash-playbook/pull/1299
- [x] wg-easy
- [x] Wordpress
- [x] writefreely
- [x] YOURLS
- [ ] Addy.io
- [ ] Any XMPP Server like ejabberd
- [ ] Bluesky
- [ ] Bonfire
- [ ] Coturn
- [ ] CrowdSec
- [ ] Discourse
- [ ] docker-doh
- [ ] Docs
[...]

PR #230, opened 2024-07-09
Add Writefreely service

ZenKyma:
Following #12, I finally managed to create a writefreely role.

Please feel free to comment and add requests!

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -88,3 +88,4 @@
 * Wg Easy: 12
 * Woodpecker Ci Agent: v2.6.0
 * Woodpecker Ci Server: v2.6.0
+* writefreely: v0.15.0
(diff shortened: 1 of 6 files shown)


spantaleev:
I've improved the documentation a bit and this is ready to merge.

One of things I've done is mention that WriteFreely can make use of a MariaDB database, but is currently not configured to do so. I think it would be preferable if support for MariaDB (and automatic wiring with the MariaDB role via `templates/group_vars_mash_servers`) is added.

I suppose WriteFreely uses SQLite when not configured to use MariaDB.

Most other services which can use a database (mostly Postgres, but sometimes MariaDB) are developed in a way to make use of it. If you're feeling like adding MariaDB support in a future PR, that'd be great!

ZenKyma:
Thank you @spantaleev ! I was indeed considering to add support for MariaDB sometime in the future when I have the time, but probably I will need some help in connecting all the dots with that. I'll open a draft pr as soon as I start working on it.

PR #231, opened 2024-07-12
Update docs to reflect changes in https://github.com/mother-of-all-se…

hooger:
…lf-hosting/ansible-role-nextcloud/pull/20


hooger:
Can you check if this can be merged? NC PR is pulled and checked, I also updated the documentation to reflect on it.

docs/services/nextcloud.md
```diff
@@ -273,7 +273,7 @@ You can also create new documents via the "plus" button.
 
 It is possible to setup preview generation, using this playbook.
 
-First modify your `vars.yml` file by adding at least the following line (other options are also present, check the corresponding `defaults/main.yml` file):
+First modify your `vars.yml` file by adding at least the following line (other options are also present, see below):
```
spantaleev, line 276: I think the previous message (mentioning defaults/main.yml) was more useful as there are many other variables (not documented below) that users may find useful

hooger: Good point, I reverted back that change. What do you think of its current form?

PR #256, opened 2024-09-11
Add Readeck service

lingawakad:
Readeck is a simple web application that lets you save the precious readable content of web pages you like and want to keep forever.
See it as a bookmark manager and a read later tool.

https://readeck.org
https://codeberg.org/readeck/readeck

please let me know if there are any issues, i am also running ```matrix-docker-ansible-deploy``` but that shouldn't affect anything hopefully?

### DIFF
--- a/docs/services/readeck.md
+++ b/docs/services/readeck.md
@@ -1,0 +1,42 @@
+# Readeck
+
+[Readeck](https://readeck.org) is a simple web application that lets you save the precious readable content of web pages you like and want to keep forever.
+See it as a bookmark manager and a read later tool.
+
+
+## Dependencies
+
+This service requires the following other services:
+
+- a [Traefik](traefik.md) reverse-proxy server
+
+
+## Configuration
+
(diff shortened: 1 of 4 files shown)


spantaleev:
Is it not possible to apply some additional security restrictions to the `docker run`-started Readeck container? Like:

- `--user=UID:GID`, to run as a regular user
- `--read-only`, to prevent file writes to the container's filesystem
- `--cap-drop=ALL` (and potentially whitelisting some capabilities, if necessary)

------

`templates/systemd/readeck.service.j2` says:

> {# We mount a tmpfs at /tmp to allow readeck migrations to run. #}

.. but given that this container's filesystem is not read-only and should be allowing writes to `/tmp` anyway, I'm guessing that this comment is a leftover from some other role. Perhaps this part should be removed, unless you find it to be necessary if you try applying `--read-only`.

lingawakad:
Added - good catch, thanks. not sure how i left those off...

PR #257, opened 2024-09-11
Add Forgejo Runner service

sergiodj:
Add support for [Forgejo Runner](https://code.forgejo.org/forgejo/runner).

Caveats:

- The software is still alpha, and not considered secure enough to be deployed in production.  I left a warning on the documentation.

- Unfortunately it's necessary to run the container as `root`, because upstream's Dockerfile will forcefully set `1000:1000` as the `UID/GID` and we need access to `/var/run/docker.sock`.

- Maybe this should use `docker-socket-proxy`?  Not sure how to integrate both.

- The registration process is not as streamlined as with [Gitea's runner](https://gitea.com/gitea/act_runner).  I had to do some `ExecPreStart` dance in the systemd service.  Hopefully that's not too problematic.


spantaleev:
Looks good! Thank you!

I think container-socket-proxy may have been used, but since it's already used by Traefik (and is configured for what Traefik needs by default), it may be more difficult to adjust (if adjustments are necessary).

If you'd like to look into improving security by integrating container-socket-proxy, that'd be nice! But I guess we're fine even as it is now.

PR #267, opened 2024-09-23
Update stirling-pdf

Bergruebe:
This new version resolves also the bug mentioned in the matrix room.


spantaleev:
Thanks!

We recommend tagging with the original version number and a suffix (e.g. `-0`, `-1`) that matches your "release" of this version.

This way, you can push fixes related to the same upstream version (`v0.29.0`) and tag each one separately. Consider adopting this versioning scheme for your future releases.

PR #274, opened 2024-10-02
Add hubsite configuration for Navidrome

Kuchenmampfer:
I noticed that Navidrome is missing on my hubsite.

This change does not fix that for me, but I think it is part of the fix and I could not find any other obvious thing. So I hope you can point me to what else needs to be changed or add it to this PR yourself?

### DIFF
--- a/templates/group_vars_mash_servers
+++ b/templates/group_vars_mash_servers
@@ -3055,6 +3055,19 @@ mash_playbook_hubsite_service_list_auto_itemized:
     }}
   # /role-specific:languagetool
 
+  # role-specific:linkding
+  - |-
+    {{
+      ({
+        'name': hubsite_service_linkding_name,
+        'url': hubsite_service_linkding_url,
+        'logo_location': hubsite_service_linkding_logo_location,
+        'description': hubsite_service_linkding_description,
+        'priority': hubsite_service_linkding_priority,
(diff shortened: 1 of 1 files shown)


spantaleev:
Thanks!

Other required parts are:

- actually injecting these new variables into `mash_playbook_hubsite_service_list_auto_itemized` in `templates/group_vars_mash_servers` ( a bit below)
- adding the Navidrome logo (`navidrome.png`) to the [Hubsite role](https://github.com/mother-of-all-self-hosting/ansible-role-hubsite), under `assets/` - a separate PR

Kuchenmampfer:
Thanks for the pointer :)

I have injected the variables and the logo was already available. So I think this should be ready now.

However, when I tested it, I got a 404 error until I restarted the `mash-hubsite` container manually. I have no idea what causes this as all the html and css files were present and I could not get logs from that container. So maybe you want to test for this yourself before merging I guess...

spantaleev:
I suppose Hubsite needs to be reinstalled and restarted as well before these logo changes can take effect.

--------

I think we're still missing a PR to the Hubsite role before merging this.

Kuchenmampfer:
no, that role already has the logo

spantaleev:
Oh, you're right! Merged now! Thanks! :bow:

PR #275, opened 2024-10-03
make proposed hostname and path consistent with other services

Kuchenmampfer:


spantaleev:
Has Striling PDF been confirmed to work correctly under a path prefix?

Kuchenmampfer:
Oh is that unclear?

I got it working yesterday, but now I get a 404 error. If that is the reason, I will change this PR to add a hint that a prefix might not work.

spantaleev:
I don't know. I don't personally use this service and have never tested it.

Kuchenmampfer:
Alright. I don't need it desperately enough to justify throwing another subdomain through acme and my reverse proxies just for testing if that fixes it right now. So I guess we can leave the docs as they are and I close this PR.

Kuchenmampfer:
I can now confirm, that stirling-pdf does work under a prefix-path, but that it binds traefik for the root path and thus messes up anything else that needs the root path. So keeping it under another subdomain in the example is indeed the right choice.

PR #300, opened 2024-11-13
Update stirling-pdf

Kuchenmampfer:


templates/requirements.yml
```diff
@@ -356,7 +356,7 @@
   name: ssh
   activation_prefix: system_security_ssh_
 - src: git+https://github.com/Bergruebe/ansible-role-stirling-pdf.git
-  version: v0.29.0
+  version: v0.32.0
```
spantaleev, line 359: There's no `v0.32.0` tag in https://github.com/Bergruebe/ansible-role-stirling-pdf, so this will not work.

Kuchenmampfer: Alright, I also opened a PR there: https://github.com/Bergruebe/ansible-role-stirling-pdf/pull/2

Kuchenmampfer:
since we are already at v0.33.0, I will close this

PR #317, opened 2024-12-12
Add SearXNG service

sergiodj:
This PR adds a new service for SearXNG.

Unfortunately, upstream doesn't version their images (and it seems like they've stopped versioning their regular software releases as well), so I decided to make the service always pull the `latest` tag.  I realize this is not ideal and am open to suggestions on how to approach this problem.


neokyx:
Great addition, thanks !

PR #340, opened 2025-02-05
Add Radarr support

spatterIight:
This PR adds Radarr support. It has been tested as working. `radarr_path_prefix` has not been tested.

https://github.com/spatterIight/ansible-role-radarr/tree/main

#### Other information:

The role configures Radarr with security in mind by doing the following:

1. Running the container as a non-root user
2. Making the filesystem read-only
3. Dropping most capabilities

Unfortunately, due to upstream requirements, some admissions had to be made:

1. Several capabilities related to permissions are added to the container
   - SETUID
   - SETGID
   - CHOWN
   - FOWNER
   - DAC_OVERRIDE
2. A `tmpfs` volume is mounted with `exec` permissions

You can read more about these upstream requirements in the documentation:

1. https://docs.linuxserver.io/misc/non-root/
2. https://docs.linuxserver.io/misc/read-only/


spantaleev:
Is this an alternative to Sonarr or a companion application?

They seem related. It's probably best if this relationship is explained in both docs pages and cross-links between said pages is created.

spatterIight:
> Is this an alternative to Sonarr or a companion application?
> 
> They seem related. It's probably best if this relationship is explained in both docs pages and cross-links between said pages is created.

A sister application, basically Sonarr works exclusively with TV shows and Radarr works exclusively with movies.

They do not work with each other, but are likely to work with other sister applications I'll be adding in the future like Jackett, Overseerr, and qBittorrent.

Commonly this is referred to as a "*Arr Stack" or something like that. Will see about adding that documentation to the Sonarr and Radarr pages.

spatterIight:
I added a note linking Sonarr and Radarr to each other. I am not sure if adding more than the simple note that is there is a good idea since it will probably have to be re-written once more *Arr applications are supported

PR #352, opened 2025-02-17
Add I hate money

IUCCA:
This PR adds the [ansible-role](https://github.com/IUCCA/ansible-role-ihatemoney) for [i hate money](https://github.com/spiral-project/ihatemoney). 
I hope that this time I have thought of all the places where a new service needs to be added.

### DIFF
--- a/templates/group_vars_mash_servers
+++ b/templates/group_vars_mash_servers
@@ -656,6 +656,11 @@ mash_playbook_devture_systemd_service_manager_services_list_auto_itemized:
     {{ ({'name': (semaphore_identifier + '.service'), 'priority': 2000, 'groups': ['mash', 'semaphore']} if semaphore_enabled else omit) }}
   # /role-specific:semaphore
 
+  # role-specific:send
+  - |-
+    {{ ({'name': (send_identifier + '.service'), 'priority': 2000, 'groups': ['mash', 'send']} if send_enabled else omit) }}
+  # /role-specific:send
+
   # role-specific:soft_serve
   - |-
     {{ ({'name': (soft_serve_identifier + '.service'), 'priority': 2000, 'groups': ['mash', 'soft-serve']} if soft_serve_enabled else omit) }}
@@ -3070,6 +3075,16 @@ hubsite_service_radicale_description: "Sync contacts and calendars"
 hubsite_service_radicale_priority: 1000
 # /role-specific:radicale
 
+# role-specific:send
+# Send
+hubsite_service_send_enabled: "{{ send_enabled }}"
+hubsite_service_send_name: Send
+hubsite_service_send_url: "https://{{ send_hostname }}{{ send_path_prefix }}"
+hubsite_service_send_logo_location: ""
+hubsite_service_send_description: "Send end-to-end encrypted files to others with a link."
+hubsite_service_send_priority: 1000
+# /role-specific:send
+
 # role-specific:syncthing
 # Syncthing
 hubsite_service_syncthing_enabled: "{{ syncthing_enabled }}"
@@ -3496,6 +3511,19 @@ mash_playbook_hubsite_service_list_auto_itemized:
       } if hubsite_service_uptime_kuma_enabled else omit)
     }}
   # /role-specific:uptime_kuma
+
+  # role-specific:send
+  - |-
+    {{
+      ({
+        'name': hubsite_service_send_name,
+        'url': hubsite_service_send_url,
+        'logo_location': hubsite_service_send_logo_location,
+        'description': hubsite_service_send_description,
+        'priority': hubsite_service_send_priority,
+      } if hubsite_service_send_enabled else omit)
+    }}
+  # /role-specific:send
 
   # role-specific:syncthing
   - |-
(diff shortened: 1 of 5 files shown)


IUCCA:
No, the container writes the environment variables in /etc/ihatemoney/ihatemoney.cfg and also requires a writable /tmp.

Unfortunately, the container does not work rootless either, as the container writes the gid to /etc/group at the first start if you want to use an alternative gid

IUCCA:
I have now experimented a bit with read-only and that works when /mash/ihatemoney/data is mounted to /etc/ihatemoney/. However, this assumes that /mash/ihatemoney/data belongs to root:root. The owner mash:mash does not work, even if you pass the ids via env file.

I have implemented this in the tag v6.1.5-1. 

Or tag v6.1.5-0 without read-only and without mount to /etc/ihatemoney, then there is no file with the owner root:root in /mash/ihatemoney.

docs/services/ihatemoney.md
```diff
@@ -0,0 +1,46 @@
...
+# ihatemoney                                                              #
+#                                                                      #
+########################################################################
+
+ihatemoney_enabled: true
+
+# To enable the Admin dashboard, first generate a hashed password with:
+# docker run -it --rm --entrypoint ihatemoney ihatemoney/ihatemoney generate_password_hash
+#ihatemoney_admin_password:
+ihatemoney_environment_variable_public: false
```
spantaleev, line 29: It's a bit unclear what `ihatemoney_environment_variable_public` actually does.

Should it be set to `true` at some point later?

docs/services/ihatemoney.md
```diff
@@ -0,0 +1,46 @@
...
+########################################################################
+#                                                                      #
+# ihatemoney                                                              #
+#                                                                      #
+########################################################################
+
+ihatemoney_enabled: true
+
+# To enable the Admin dashboard, first generate a hashed password with:
+# docker run -it --rm --entrypoint ihatemoney ihatemoney/ihatemoney generate_password_hash
```
spantaleev, line 27: Maybe we can also suggest a `docker exec` command to run on the server once the service is up and running?

.. for folks that would prefer to do that.

templates/group_vars_mash_servers
```diff
@@ -3324,6 +3363,70 @@ hubsite_service_list_auto: "{{ mash_playbook_hubsite_service_list_auto_itemized
...
+
+# role-specific:exim_relay
+ihatemoney_email_host: "{{ exim_relay_identifier if exim_relay_enabled else '' }}"
+ihatemoney_email_port: "{{ 8025 if exim_relay_enabled else '' }}"
+ihatemoney_default_from_email: "{{ exim_relay_sender_address if exim_relay_enabled else '' }}"
+# /role-specific:exim_relay
+
+########################################################################
+#                                                                      #
+# /ihatemoney                                                             #
```
spantaleev, line 3424:
```suggestion
# /ihatemoney                                                          #
```

templates/group_vars_mash_servers
```diff
@@ -3324,6 +3363,70 @@ hubsite_service_list_auto: "{{ mash_playbook_hubsite_service_list_auto_itemized
 # /role-specific:hubsite
 
 
+# role-specific:ihatemoney
+########################################################################
+#                                                                      #
+# ihatemoney                                                              #
```
spantaleev, line 3369:
```suggestion
# ihatemoney                                                           #
```

docs/services/ihatemoney.md
```diff
@@ -0,0 +1,46 @@
...
+#ihatemoney_admin_password:
+ihatemoney_environment_variable_public: false
+ihatemoney_hostname: mash.example.com
+ihatemoney_path_prefix: /ihatemoney
+
+########################################################################
+#                                                                      #
+# /ihatemoney                                                             #
```
spantaleev, line 35:
```suggestion
# /ihatemoney                                                          #
```

docs/services/ihatemoney.md
```diff
@@ -0,0 +1,46 @@
...
+
+
+## Configuration
+
+To enable this service, add the following configuration to your `vars.yml` file and re-run the [installation](../installing.md) process:
+
+```yaml
+########################################################################
+#                                                                      #
+# ihatemoney                                                              #
```
spantaleev, line 20:
```suggestion
# ihatemoney                                                           #
```

docs/services/ihatemoney.md
IUCCA: I have changed this to docker exec

docs/services/ihatemoney.md
IUCCA: If your instance is meant to be open for everyone to freely create projects, you would set it to true.
If your instance is intended for private use, and you want to restrict project creation to users with the admin password, you would set it to false (or simply not set it, as false is the default).

I have adapted the documentation a little, now it should be clearer

spatterIight:
There are several Ansible variables defined in the role that do not seem to be used:

1. `ihatemoney_scheme`
2. `ihatemoney_systemd_wanted_services_list`
3. `ihatemoney_container_image_force_pull`
4. `ihatemoney_container_additional_mounts`
5. `ihatemoney_container_labels_traefik_http_middlewares`

These are all defined in `defaults/main.yml` but not actually used anywhere in the role.

spatterIight:
These lines in the role confuse me:

```
# Controls the ihatemoney environment debug variable
ihatemoney_environment_variable_ihatemoney_port: 8000
```

https://github.com/IUCCA/ansible-role-ihatemoney/blob/36f63e48c0983a2122487309fce08ebcfda5c948/defaults/main.yml#L131C1-L132C54

Is the comment a typo?

spatterIight:
In the role two environment variables are commented out:

```
#PUID={{ ihatemoney_uid }}
#PGID={{ ihatemoney_gid }}
```

Are these supposed to be commented out? If they don't do anything it may be better to just remove them

https://github.com/IUCCA/ansible-role-ihatemoney/blob/36f63e48c0983a2122487309fce08ebcfda5c948/templates/env.j2#L9

spantaleev:
Maybe [my review comment for DokuWiki's PR](https://github.com/mother-of-all-self-hosting/mash-playbook/pull/386#issuecomment-2695544917) could also be useful for improving your ihatemoney Ansible role as well.

It seems like:

- your "Ensure ihatemoney container image is pulled" task is different than the ones we usually use for other roles and does not respect `devture_playbook_help_container_retries_count`, etc.

- your "Ensure ihatemoney container network is created" task does not respect `devture_systemd_docker_base_ipv6_enabled` and `devture_systemd_docker_base_container_networks_driver_options`, etc.

- there's a commented-out line (`#    - "{{ ihatemoney_data_path }}"`) in `tasks/install.yml`

- there are some empty newlines in `tasks/install.yml`

- `docker.service is hardcodedin `ihatemoney_systemd_required_services_list_default: ['docker.service']` in `defaults/main.yml`, instead of using `devture_systemd_docker_base_docker_service_name`

Please see that other review comment (for DokuWiki) for more details and pick what's relevant for improving your ihatemoney role (and any other roles you maintain that you'd like to improve with new ideas/solutions that we adopt throughout mash-playbook).

PR #354, opened 2025-02-18
Update requirements.yml

lingawakad:
Apparently something changed and CW does _not_ like headers anymore


lingawakad:
scratch that, was just a temporary fix, seems to be more related to proxying

PR #378, opened 2025-02-24
Add LabelStudio support to the MASH repo

hooger:


docs/services/labelstudio.md
```diff
@@ -0,0 +1,61 @@
...
+#                                                                      #
+# LabelStudio                                                          #
+#                                                                      #
+########################################################################
+
+labelstudio_enabled: true
+labelstudio_hostname: "example.labelstudio.hostname.com"
+
+# Although this is not strictly necessary, it is highly recommended
+labelstudio_postgre_backend_enabled: true
```
spatterIight, line 28: typo?

docs/services/labelstudio.md
```diff
@@ -0,0 +1,61 @@
...
+labelstudio_environment_variables_username: "admin-username"
+labelstudio_environment_variables_password: "admin-user-password"
+```
+
+
+## Recommended other services
+
+It is possible to attach a pre-labeling backend to LabelStudio.
+One such example project can be found in [this repository](https://github.com/seblful/label-studio-yolo-backend).
+Although we created a role for such backends in-house, currently it is not in a state to be upstreamed as it contains too many specific variables.
```
spatterIight, line 61: Who is "we"? The reader may not understand and think "we" = the MASH github organization

templates/requirements.yml
```diff
@@ -184,6 +184,10 @@
   version: v6.3.4-3
   name: keydb
   activation_prefix: keydb_
+- src: git+https://gitlab.com/horvathg.1988/ansible-role-labelstudio.git
+  version: hg-backend
```
spatterIight, line 188: usually the role version is a specific version tag, in the format:

v1.0.0-0

Where `v` is always prefixed for version, `1.0.0` is the actual version of the application being deployed (v1 of label studio) and the `-0` indicates the role revision.

docs/services/labelstudio.md
```diff
@@ -0,0 +1,61 @@
...
+# LabelStudio                                                          #
+#                                                                      #
+########################################################################
+
+labelstudio_enabled: true
+labelstudio_hostname: "example.labelstudio.hostname.com"
+
+# Although this is not strictly necessary, it is highly recommended
+labelstudio_postgre_backend_enabled: true
+labelstudio_database_password: "some-very-serious-password"
```
spatterIight, line 29: Since a MASH postgres server is listed as a dependency of this service, why does this variable have to be defined?

I am not an expert on this, but couldn't it use some variable that is already defined?

For example, the Authentik and Nextcloud roles do the following:

```yaml
authentik_database_password: "{{ '%s' | format(mash_playbook_generic_secret_key) | password_hash('sha512', 'db.authentik', rounds=655555) | to_uuid }}"

nextcloud_database_password: "{{ '%s' | format(mash_playbook_generic_secret_key) | password_hash('sha512', 'db.nextcloud', rounds=655555) | to_uuid }}"
```

spatterIight:
There is broken J2 in the role at these lines:

1. https://gitlab.com/horvathg.1988/ansible-role-labelstudio/-/blob/hg-backend/templates/env.j2?ref_type=heads#L15
2. https://gitlab.com/horvathg.1988/ansible-role-labelstudio/-/blob/hg-backend/templates/env.j2?ref_type=heads#L16

It looks like you are just missing `{{ }}`

docs/services/labelstudio.md
spatterIight: it looks like this is actually not really a typo.. kind of..

the actual upstream LabelStudio project uses 'POSTGRE' instead of 'POSTGRES' in their various naming for god knows why

https://github.com/HumanSignal/label-studio/blob/05535fe86d9dc031f2f81857ebdbe981da4d1864/docs/source/guide/storedata.md?plain=1#L39

templates/requirements.yml
hooger: https://hub.docker.com/r/heartexlabs/label-studio they suggest using the latest tag and IIUC they are naming their current docker images based on commits. What should I input into the `labelstudio_version` variable and also how should I tag the version?

docs/services/labelstudio.md
hooger: Fixed and unified to `postgres`

docs/services/labelstudio.md
hooger: Removed the sentence. Fixed

hooger:
Fixed the missing variable handlers  ̇{{ }}` and also removed the unused variable line. Thanks for spotting those as well!

spatterIight:
> Fixed the missing variable handlers ̇{{ }}` and also removed the unused variable line. Thanks for spotting those as well!

Thanks! Were you able to push those changes to the role? When I go to the branch defined in `requirements.yml` I still see the same issue

templates/requirements.yml
spatterIight: There are some roles that pin their default version to latest, so there is some precedence here. For example the `searxng` role is set-up as follows:

```
- src: git+https://git.sergiodj.net/sergiodj/ansible-role-searxng.git
  version: v1.0-0
  name: searxng
```
https://github.com/mother-of-all-self-hosting/mash-playbook/blob/732973f7c70d4adf478728d752b2dcc6b92f051f/templates/requirements.yml#L366C1-L368C16

```
# Upstream doesn't release versioned images, so we have to stick with
# "latest" here.
searxng_version: latest
```
https://git.sergiodj.net/sergiodj/ansible-role-searxng/src/commit/b06e4e9d032741d7d29ac92e0ac98c5ca8bc8688/defaults/main.yml#L11

So in this case I would probably tag the role as `v1.0.0` and add a comment about why `latest` is the default version in the role

Thanks!

docs/services/labelstudio.md
hooger: And also reverted so labelstudio still functions

hooger:
I think the current version is coherent. Please tell me if you have any more comments!

spatterIight:
https://gitlab.com/horvathg.1988/ansible-role-labelstudio/-/blob/v1.0.0-2/meta/main.yml?ref_type=tags#L2

In the role it mentions the author/company/namespace is `etke.cc`. I think this is not correct.

Company and namespace can simply be removed, not sure if author is a hard requirement. In my role I set author to me username

https://github.com/spatterIight/ansible-role-sonarr/blob/b8f663be63c3f87b80daaeb50e24592c1f018b1f/meta/main.yml#L3

spatterIight:
I am not an official maintainer of MASH, so I do not have permission to officially approve/merge this PR, but I think everything is looking pretty good

hooger:
Nice catch, fixed the meta-data as well. Thanks for the review!

PR #385, opened 2025-03-03
Update configuring-playbook.md: add an instruction to create a git repository for vars.yml

luixxiul:


luixxiul:
That indeed makes much sense ... I've managed my repository for MDAD project editing the `.gitignore` file :rofl:

PR #395, opened 2025-03-12
Add ntfy

luixxiul:
This PR intends to add ntfy, based on the existing implementation at matrix-docker-ansible-deploy.

ntfy is not just a humble UnifiedPush service but also has its own powerful functions like sending/receiving notifications with attachment, forwarding notifications to email address, and so on.

As this is my first ever trial of implementing a role to a playbook, I have to admit that there is likely to be misconfiguration on `templates/group_vars_mash_servers`… Please feel free to correct them as they should be.


spantaleev:
Looks good! :bow:

luixxiul:
Thanks for reviewing!

PR #400, opened 2025-03-13
Add Cryptpad support

luixxiul:
This does not work yet: see https://github.com/mother-of-all-self-hosting/ansible-role-cryptpad/pull/7.

Any helps are appreciated!

### DIFF
--- a/docs/services/cryptpad.md
+++ b/docs/services/cryptpad.md
@@ -1,10 +1,10 @@
 <!--
-SPDX-FileCopyrightText: 2020 - 2024 MDAD project contributors
-SPDX-FileCopyrightText: 2020 - 2024 Slavi Pantaleev
 SPDX-FileCopyrightText: 2020 Aaron Raimist
 SPDX-FileCopyrightText: 2020 Chris van Dijk
 SPDX-FileCopyrightText: 2020 Dominik Zajac
 SPDX-FileCopyrightText: 2020 Mickaël Cornière
+SPDX-FileCopyrightText: 2020-2024 MDAD project contributors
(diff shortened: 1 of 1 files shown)


moan0s:
Hi @luixxiul,
I fixed the role and tested it successfully. Maybe you can check as well? I also added some basic documentation.

Basically, the issue was, that [Cryptpad needs one certificate, that covers both domains](https://docs.cryptpad.org/en/admin_guide/installation.html#admin-domain-config). Back when I created that role, I didn't know how to do this, however I now found a solution (see 63d6bc91199c18745c79b9d1af80a83ba2c51bfb).

luixxiul:
I'll check and get back to you soon!

luixxiul:
Seemingly its UI works, but `/checkup/` page (cryptpad.example.com/checkup/) returns three critical errors related to Content-Security-Policy headers.

moan0s:
> Seemingly its UI works, but /checkup/ page (cryptpad.example.com/checkup/) returns three critical errors related to Content-Security-Policy headers. 

I tried various things to fix it, so far I'm not successful. any help is appreciated. I'll also continue to try.

moan0s:
FYI: I run my instance since my last comment. The CSP errors seem to "only" affect security, not stability/function

luixxiul:
I've finally managed to let the tests related to the CSP headers passed — basically they are specified by the service behind Traefik and overriding them with ours has stopped from them being applied properly.

I have confirmed that all tests passed (53 / 53 tests passed) with my own VPS server and it should work without compromising security.

PR #401, opened 2025-03-13
Replace hyphens with genuine em dashes

luixxiul:
This might be worth reading: https://www.merriam-webster.com/grammar/em-dash-en-dash-how-to-use (on how em dash, en dash, and hyphen should be used)

### DIFF
--- a/docs/developer-documentation.md
+++ b/docs/developer-documentation.md
@@ -54,7 +54,7 @@ There are a few files that you need to adapt:
 <details>
 
 <summary> file: templates / group_vars_mash_servers </summary>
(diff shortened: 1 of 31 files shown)


spantaleev:
Makes sense, but.. I think the issue is that:

- most people can't easily enter these characters on their keyboards or it's a hassle to do it
- most people don't know what the difference is
- or even if they know, they don't really care, since it's all pretty much the same. 3 characters that look almost the same, to the point where most people's eyes wouldn't be able to distinguish at least 2 of them (`–` vs `-` or `—` vs `–`) (depending on the font, people's familiarity with the font, etc.), should probably not even exist and we should never have to think about this ;)

That said, thank you for making our docs use the proper type of dash/hyphen!

PR #416, opened 2025-03-22
Update funkwhale.md

luixxiul:
This PR intends to update funkwhale,md especially focusing on how to set up a Valkey instance, based on my experience to implement a new role to the playbook, following this documentation as an example.

Since installing a dedicated Valkey instance is the step which you will not often encounter, I think it needs detailed explanation about why it is needed, how it can be installed, etc. on the same page, in order to spare readers taking time to study the documentation.

### DIFF
--- a/docs/services/funkwhale.md
+++ b/docs/services/funkwhale.md
@@ -1,6 +1,7 @@
 <!--
 SPDX-FileCopyrightText: 2023 Julian-Samuel Gebühr
 SPDX-FileCopyrightText: 2023-2024 Slavi Pantaleev
+SPDX-FileCopyrightText: 2025 Suguru Hirahara
 
 SPDX-License-Identifier: AGPL-3.0-or-later
 -->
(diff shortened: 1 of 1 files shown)


luixxiul:
This PR is ready for review.

Hopefully this update should also be applicable to other documentation files of services which need Valkey.

spantaleev:
Sure, it looks good! I've always found it a bit annoying that the same Valkey-related information needs to be duplicated everywhere.. And that it waters down the documentation you're actually trying to read.

But.. maybe that's best.

luixxiul:
I thought that as well, but the thing is that the way of naming variables is not strictly same among the services.

For example, for Funkwhale we've got `funkwhale_api_systemd_required_services_list_custom` (mind `_api_`) while there is `peertube_systemd_required_services_list_custom` for Peertube, so simply specifying  `SERVICE_systemd_required_services_list_custom` on a general docs (like there is a common guide for mautrix bridges on the MDAD playbook) does not work.

Once we've got applied the common pattern to find out what's actually common what's not, we can revisit this topic :wink:

PR #434, opened 2025-03-29
Add ihatemoney

IUCCA:
I accidentally deleted the first  PR #352  so I'm reposting it now. I have implemented the suggestions from the DokuWiki PR. 

@spatterIight  and @spantaleev , thanks for your feedback - I hope the role is satisfactory now, otherwise I'll continue to rework it.

Ignore the version change of Tandoor - it has already been updated by the PR #433.


spantaleev:
Thank you! :bow:

PR #474, opened 2025-04-20
Add Matomo

olli-acterglobal:
This PR adds [Matomo](https://matomo.org/), a self-hosted analytics service. 
I wrote a role for this purpose: https://github.com/acterglobal/ansible-role-matomo

This PR should be complete and working, but it's my first MASH PR so comments welcome :)

### DIFF
--- a/docs/services/matomo.md
+++ b/docs/services/matomo.md
@@ -23,6 +23,13 @@ To enable this service, add the following configuration to your `vars.yml` file 
 matomo_enabled: true
 
 matomo_hostname: matomo.example.com
+
+########################################################################
+#                                                                      #
+# /matomo                                                              #
+#                                                                      #
+########################################################################
+
 ```
 
 ## Usage
--- a/docs/services/matomo.md
+++ b/docs/services/matomo.md
@@ -1,0 +1,37 @@
+# Matomo
+
+[Matomo](https://matomo.org/) (formerly Piwik) is a free and open source web analytics platform.
+
+## Dependencies
+
+This service requires the following other services:
+
+- [MariaDB](mariadb.md) database
+- [Traefik](traefik.md) reverse-proxy server
+
+## Configuration
+
+To enable this service, add the following configuration to your `vars.yml` file and re-run the [installation](../installing.md) process:
+
+```yaml
+########################################################################
+#                                                                      #
+# matomo                                                               #
+#                                                                      #
+########################################################################
+
+matomo_enabled: true
+
+matomo_hostname: matomo.example.com
+```
+## Usage
+
+After installation, visit your Matomo instance at the configured URL. You'll be guided through a setup wizard where you can:
+
+1. Create your administrator account
+2. Configure your first website to track
+3. Get the tracking code to add to your website
+
+Add the tracking code to your website's HTML to start collecting analytics data.
+
+For additional configuration options, refer to [ansible-role-matomo](https://github.com/mother-of-all-self-hosting/ansible-role-matomo)'s `defaults/main.yml` file.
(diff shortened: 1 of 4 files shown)


templates/group_vars_mash_servers
```diff
@@ -4702,6 +4718,63 @@ mariadb_managed_databases_auto: "{{ mash_playbook_mariadb_managed_databases_auto
 
 
 
+# role-specific:matomo
+########################################################################
+#                                                                      #
+# matomo                                                                #
```
luixxiul, line 4724:
```suggestion
# matomo                                                               #
```

templates/requirements.yml
```diff
@@ -503,3 +503,7 @@
   version: v1.10.0-0
   name: yourls
   activation_prefix: yourls_
+- src: git+https://github.com/acterglobal/ansible-role-matomo.git
+  version: main
+  name: matomo
+  activation_prefix: matomo_
```
luixxiul, line 509: I'd add this block to just above `git+https://github.com/mother-of-all-self-hosting/ansible-role-matterbridge.git` (alphabetically).

templates/setup.yml
```diff
@@ -527,6 +527,10 @@
     - role: galaxy/matterbridge
     # /role-specific:matterbridge
 
+    # role-specific:matomo
+    - role: galaxy/matomo
+    # /role-specific:matomo
+
```
luixxiul, line 533: I'd move this above as well — the block for matterbridge should have not been added there in the first place…

olli-acterglobal:
Beware reviewers - I just force pushed to this branch to fix the committer name. No other changes.

Thank you for the review @luixxiul, I have implemented your changes :).

spantaleev:
Thank you for working on Matomo support!

Pinning the role to some tag (like we do for all other roles) would be better than using `main`.

By the way, your role is a bit outdated in the way it creates the Matomo container network (lacks IPv6 and other driver options that other roles use). See [how the forgejo role does it](https://github.com/mother-of-all-self-hosting/ansible-role-forgejo/blob/68ede5ac426f8a652db20ea496a14498fc0bd036/tasks/install.yml#L37-L42) for a more up-to-date example.

docs/services/matomo.md
```diff
@@ -0,0 +1,38 @@
...
+########################################################################
+#                                                                      #
+# matomo                                                               #
+#                                                                      #
+########################################################################
+
+matomo_enabled: true
+
+matomo_hostname: matomo.example.com
+```
```
luixxiul, line 26: Did the end block disappear for some reason?

olli-acterglobal:
> Pinning the role to some tag (like we do for all other roles) would be better than using `main`.

Thank you, fixed.

> By the way, your role is a bit outdated in the way it creates the Matomo container network (lacks IPv6 and other driver options that other roles use). See [how the forgejo role does it](https://github.com/mother-of-all-self-hosting/ansible-role-forgejo/blob/68ede5ac426f8a652db20ea496a14498fc0bd036/tasks/install.yml#L37-L42) for a more up-to-date example.

Excellent! Done. I also see that developer-documentation.md is already updated with new examples :)

spantaleev:
The `# Project source code URL:` reference in the role's `defaults/main.yml` should point to the upstream sources: https://github.com/matomo-org/matomo

`just opml` in the playbook then can correctly find them and generate a RSS/Atom feed for finding "new releases".

Please update that and tag a new version and let's merge this!

olli-acterglobal:
> Please update that and tag a new version and let's merge this!

Alright, all done! Thanks for the walkthrough :)

spantaleev:
Thanks! :bow:

PR #479, opened 2025-04-23
Upgrade Send (latest → 3.4.24)

luixxiul:
Containers are available at https://gitlab.com/timvisee/send/container_registry/1418504.


spantaleev:
By the way, `v2025.2.4-1` seems like a newer (bigger) version number than `v3.4.24-0`, so `agru -u` may do the wrong thing and try to "upgrade" back to `v2025.2.4-1`.

Perhaps getting rid of the `v2025*` tags in the role's repository would be the simplest workaround.

PR #521, opened 2025-05-04
Added minecraft role

XHawk87:


docs/services/minecraft.md
```diff
@@ -0,0 +1,90 @@
...
+  INITIAL_ENABLED_PACKS=vanilla,bundle
+
+########################################################################
+#                                                                      #
+# /minecraft                                                           #
+#                                                                      #
+########################################################################
+```
+
+In the example configuration above, we configure the server to be hosted at `mash.example.com:25565`. It will be the latest Paper minecraft server with 2GB of RAM, hard difficulty, only you on the whitelist, rather generous render distance, PvP, bundles, and an automatic server pause when the last player leaves.
```
luixxiul, line 73:
```suggestion
In the example configuration above, we configure the server to be hosted at `mash.example.com:25565`. It will be the latest Paper Minecraft server with 2GB of RAM, hard difficulty, only you on the whitelist, rather generous render distance, PvP, bundles, and an automatic server pause when the last player leaves.
```

An anchor link for "Paper Minecraft server" would be appreciated :slightly_smiling_face:

docs/services/minecraft.md
```diff
@@ -0,0 +1,90 @@
...
+```yaml
+########################################################################
+#                                                                      #
+# minecraft                                                            #
+#                                                                      #
+########################################################################
+
+minecraft_enabled: true
+minecraft_port: 25565
+minecraft_docker_version: latest
```
spatterIight, line 32: just 'minecraft_version' might be better here, unless im missing something

docs/services/minecraft.md
spatterIight, line 73: The first sentence confuses me, where in the example configuration is the server set to be hosted at 'mash.example.com:25565'?

Would it be more accurate to say port 25565 is simply exposed and needs to be port-forwarded?

docs/services/minecraft.md
```diff
@@ -0,0 +1,90 @@
...
+
+```yaml
+########################################################################
+#                                                                      #
+# minecraft                                                            #
+#                                                                      #
+########################################################################
+
+minecraft_enabled: true
+minecraft_port: 25565
```
spatterIight, line 31: In the role the way you manage forwarding ports seems slightly non-standard
 In most other MASH roles there is a concept of a port and a bind port.

Example: https://github.com/mother-of-all-self-hosting/ansible-role-qbittorrent/blob/485651816568977c541300e0327fb470a68970fb/templates/systemd/qbittorrent.service.j2#L31

docs/services/minecraft.md
XHawk87: This is the version of the itzg/minecraft-server docker image. `minecraft_version` would imply the version of the actual Mojang Minecraft Server, which is set using `MINECRAFT_VERSION` environment variable.

templates/group_vars_mash_servers
```diff
@@ -5343,6 +5348,31 @@ matomo_container_labels_traefik_tls_certResolver: "{{ traefik_certResolver_prima
...
+#                                                                      #
+########################################################################
+
+minecraft_enabled: false
+
+minecraft_identifier: "{{ mash_playbook_service_identifier_prefix }}minecraft"
+
+minecraft_base_path: "{{ mash_playbook_base_path }}/{{ mash_playbook_service_base_directory_name_prefix }}minecraft"
+
+minecraft_uid: "{{ mash_playbook_uid }}"
```
spatterIight, line 5364: In the role you set the UID and GID default variables to 1000, this seems non-standard. In most other MASH roles the default is undefined and a task checks to make sure its defined at run time

Example:
1. https://github.com/mother-of-all-self-hosting/ansible-role-qbittorrent/blob/485651816568977c541300e0327fb470a68970fb/defaults/main.yml#L9
2. https://github.com/mother-of-all-self-hosting/ansible-role-qbittorrent/blob/485651816568977c541300e0327fb470a68970fb/tasks/validate_config.yml#L9

spatterIight:
In the role is it possible to configure it to run the container in read-only mode and/or to specify CAP drop? These features make the container more secure if they work.

Example: https://github.com/mother-of-all-self-hosting/ansible-role-mosquitto/blob/e30a2114a777beee07670a6fb9d1adf1f68a0419/templates/systemd/mosquitto.service.j2#L31

docs/services/minecraft.md
spatterIight: Gotcha, maybe adding a note about that to the documentation could be good.

usually MASH roles deploy containers which run a specific version of software. Don't think I've ever seen an image that can run different versions of software based on an environment variable like that. just my 2 cents

docs/services/minecraft.md
spatterIight: Since the feature isn't available yet it might be better to just leave it out of the documentation for now until its ready

docs/services/minecraft.md
luixxiul: I do not know detail about variants at all, but I am wondering if setting it to `2025.4.2` (at https://hub.docker.com/r/itzg/minecraft-server/tags) would not work here.

XHawk87:
> In the role is it possible to configure it to run the container in read-only mode and/or to specify CAP drop? These features make the container more secure if they work.
> 
> Example: https://github.com/mother-of-all-self-hosting/ansible-role-mosquitto/blob/e30a2114a777beee07670a6fb9d1adf1f68a0419/templates/systemd/mosquitto.service.j2#L31

I can certainly try it. I don't see why it would ever need to modify anything outside of `/data`, and I'm not sure what all the caps do but I am not sure why it'd need them either. Though it might be better if it's optional since Minecraft servers are heavily moddable through mods or Bukkit plugins, and if one of those did have a need for it it would be better if they could opt out of those, so long as they know what they're doing.

docs/services/minecraft.md
XHawk87: That's what the default is set to: https://github.com/XHawk87/ansible-role-minecraft/blob/24cfe0ae5fc8a93ed1476e0b9d96e5ac5142407a/defaults/main.yml#L21

Both the default `2025.4.2` and `latest` work. I wanted to provide an example of how to change it.

XHawk87:
Perhaps a warning variable and a task to check to go along with it would be a good idea?

E.g. 
```
minecraft_security_mode: insecure
minecraft_security_write_access: true
minecraft_security_allow_capabilities: 
  - CAP_NET_ADMIN
```

docs/services/minecraft.md
luixxiul: If specifying a version is not necessary, probably it would be better not to mention it, so that readers would not be confused. Otherwise, adding a comment near this line would be nice.

spatterIight:
> Perhaps a warning variable and a task to check to go along with it would be a good idea?
> 
> E.g.
> 
> ```
> minecraft_security_mode: insecure
> minecraft_security_write_access: true
> minecraft_security_allow_capabilities: 
>   - CAP_NET_ADMIN
> ```

sure, something like this could probably work, which would allow an operator to disable read-only mode and add capabilities if needed for some reason.

what would the `minecraft_security_mode` variable do?

XHawk87:
> what would the `minecraft_security_mode` variable do?

Defaults to `secure`. If it's not set to `insecure`, you can't use any of less secure settings, and gives an explanatory warning message to make sure the user understands the consequences of changing these settings.

docs/services/minecraft.md
```diff
@@ -0,0 +1,90 @@
...
+########################################################################
+#                                                                      #
+# minecraft                                                            #
+#                                                                      #
+########################################################################
+
+minecraft_enabled: true
+minecraft_bind_port: 25565
+minecraft_container_volumes_additional:
+  - /run/screen/S-ubuntu:/run/screen/S-{{ minecraft_identifier }}
```
spantaleev, line 33: Doesn't this need to be quoted for the variable reference to work correctly?

```suggestion
  - "/run/screen/S-ubuntu:/run/screen/S-{{ minecraft_identifier }}"
```

It would probably be helpful to add some details as to why this is necessary and how people can do it correctly.

It seems like you assume that an `ubuntu` user exists on the host system, which is only the case on Ubuntu-based distros.

Is this strictly necessary to do? What should all others do here? Should they create some other user? Should they use the `mash` user?

Why mount to `/run/screen/S-{{ minecraft_identifier }}` (why `{{ minecraft_identifier }}`)?

`docker run --volume` is generally not a great option to use if you'd like to mount existing directories into the container, because:

- it does mount correctly if the host directory already exists
- if the host directory doesn't exist, it magically auto-creates it (with `root:root` or whatever magical permissions).

It's better to use `--mount`, which will fail if you provide it with an invalid host path. This lets people discover their mistake and correct it, instead of new `root:root`-owned directories being magically created.

PR #550, opened 2025-05-12
Add immich service

### LINKED ISSUE #133: Add Immich
[Immich](https://immich.app/) is a really cool Google Photos alternative for photo backup and library management, with some ML features too. It would be really nice to have in the playbook

IUCCA:
A first draft of the [ansible-role-immich.](https://github.com/IUCCA/ansible-role-immich) Immich is only considered stable from the future version 2.0, should the PR remain a draft until then? The current version is 1.132.3.

Otherwise, the role for machine learning/hardware video transcoding is only ready for use on the CPU and with Nvidia Cuda, I haven't gotten to the other platforms that immich supports yet and I don't have the opportunity to test them.

At the moment I'm still struggling with the driver installation of the Nvidia driver, because a debian update broke it after it was running once. Would it be okay at this point to refer to the original documentation from immich, or should this also be in the service documentation?

In general, the documentation needs a few more detailed instructions.

As I don't have much time at the moment, I can't say when I'll get around to it. If anyone would like to take on this, PRs are welcome.

Closes #133


Bergruebe:
The installation works for me. Sadly I had not the time to test it more.

spantaleev:
I found a few issues with this role:

1. It manages the Postgres (with extensions) database by itself. I think it'd be preferable to either fork [mother-of-all-self-hosting/ansible-role-postgres](https://github.com/mother-of-all-self-hosting/ansible-role-postgres) or **better yet** find a way to make use of it by reconfiguring its variables (changing the image tag and other config like initdb args). That way, we'll have:
  - better separation of concerns  
  - a much more robust Postgres installation which will also support upgrading in the future (when a new "Postgres with extensions" version is released) via the existing `upgrade-postgres` tag. It's possible that the Immich/Postgres container images do (or will do) their own automatic upgrades and that specialized Ansible tasks that do this manually are not necessary.

2. It pulls Postgres from `ghcr.io/tensorchord/pgvecto-rs:pg14-v0.2.0@sha256:739cdd626151ff1f796dc95a6591b55a714f341c737e27f045019ceabf8e8c52`, which is different from what Immich's [docker-compose file](https://github.com/immich-app/immich/blob/007ba1d9efdc571c87ca1553dd28dbbfe7eeee2f/docker/docker-compose.yml#L59) currently points to (`ghcr.io/immich-app/postgres:14-vectorchord0.4.3-pgvectors0.2.0@sha256:32324a2f41df5de9efe1af166b7008c3f55646f8d0e00d9550c16c9822366b4a` as of this writing)

3. It uses [a single `env` file](https://github.com/IUCCA/ansible-role-immich/blob/ce74edeabc867f04d2b033ba05637c3e10c08fcd/templates/env.j2), which contains environment variables for various services (Immich, Postgres, ..). It's better to split this. If Postgres will be managed by another role, then this one can remain just for Immich. If possible to split per Immich service (`immich-server`, `immich-machine-learning`) so that it's clear which services needs what environment variable, then that would be even better. The official [docker-compose file](https://github.com/immich-app/immich/releases/latest/download/docker-compose.yml) does not make it obvious which environment variable is necessary for which service, so this may take some digging and trial & error.

IUCCA:
After the first draft of the role, Immich changed the database, among other things. Since Immich is still not considered stable, I am waiting with further updates until that is the case.


That is why I used a separate database for Immich without existing database management in the first draft, as the update option is currently not guaranteed.

spantaleev:
Just wanted to let you folks know that I've been making progress on refactoring the Immich role and integrating it into MASH.

The good news is that I have already gotten it working.

The bad news is that I haven't published any of my changes yet. Some things (Postgres integration) are still in flux and drastic (backward incompatible changes) are still happening.  I'll let you know when it stabilizes.

This update is just to let you know that:

- things are progressing
- there's no need to do double work on this. Please wait for my follow-up

PR #555, opened 2025-05-14
Upgrade FMD Server (v0.10.0-5 → v0.11.0-0)

luixxiul:
I managed to have the upgraded version run with UID and GID set to 1000, but due to my lack of knowledge I am not sure if it is optimal or not.

Relevant information:
- https://gitlab.com/fmd-foss/fmd-server/-/commit/61dd086eb28b26eef54503df3dd60dd44eb46b54
- https://gitlab.com/fmd-foss/fmd-server#self-hosting-with-docker-compose

Please let me know if there is another better way.


spantaleev:
Sounds good! Hopefully existing data files (owned by `mash:mash`) don't cause trouble when we run it as `1000:1000`.

I wonder if the role needs to `chown` the files recursively to fix that for existing users.

PR #779, opened 2025-07-25
Update privatebin.md

loicngr:
Hi,

Add **/** before **bin**.

Loic.


luixxiul:
Confirmed it works; thanks for spotting!

PR #821, opened 2025-07-28
Update uninstalling.md

TeamMasse:
just stop 
give the error: error: Justfile does not contain recipe `stop`.
just run-tags stop 
works


spantaleev:
`just stop` does not invoke a valid just recipe, but `just stop-all` does. I've adjusted your PR a bit.

PR #841, opened 2025-08-03
bump tandoor to 2.0.1

IUCCA:


IUCCA:
The check is already built into tasks/validate_config.yml. I just thought that an additional note in the changelog file would be useful.

PR #941, opened 2025-09-05
Add Tinyauth (ansible-role-tinyauth)

luixxiul:
This intends to add [Tinyauth](https://tinyauth.app/docs/about/), a pretty simple authentication middleware which supports TOTP, OAuth, and LDAP and does not require a database or some sort of add-ons.

### DIFF
--- a/docs/services/authelia.md
+++ b/docs/services/authelia.md
@@ -204,3 +204,4 @@ To get started, open the URL with a web browser, and log in to the portal websit
 - [authentik](authentik.md) — Open-source identity provider focused on flexibility and versatility
 - [Keycloak](keycloak.md) — Open source identity and access management solution
 - [OAuth2-Proxy](oauth2-proxy.md) — Reverse proxy and static file server that provides authentication using OpenID Connect providers (Google, GitHub, authentik, Keycloak, and others) to SSO-protect services which do not support SSO natively
+- [Tinyauth](tinyauth.md) — Simple authentication middleware that adds a login screen or OAuth with Google, Github, and any provider to your Docker services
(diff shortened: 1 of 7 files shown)


spantaleev:
Or perhaps some global variable for additional Traefik labels for services (e.g. `mash_playbook_reverse_proxyable_services_additional_labels` in `roles/mash/playbook_base/defaults/main.yml`).

The playbook can then use this variable to inject it into all services.

Here's the trouble though:

- many of the "proxyable services" supported by the playbook have their own authentication method, so while it's good to have some variable (like `mash_playbook_reverse_proxyable_services_additional_labels`) that gets injected everywhere.. in our case, we'd probably wish to have one more variable (`mash_playbook_reverse_proxyable_authenticationless_services_additional_labels`) which only targets services that lack authentication. This way, people would be able to target both *all* services (for whatever reason they may have), or *only those that lack authentication*.

- we'd need to go through all roles and make sure they provide support for adding "additional labels". Some roles have a variable for this (e.g. `syncthing_container_labels_additional_labels`). However, a few different "additional labels" variables would be better (e.g. `_auto`, for the playbook to inject stuff like `mash_playbook_reverse_proxyable_services_additional_labels` and `mash_playbook_reverse_proxyable_authenticationless_services_additional_labels`; `_custom`, in case the user wants to inject some labels into a specific service - perhaps some additional labels on top of all the others)

- we'd need to adjust the group variables to override all these `_auto` additional labels variables for all services, injecting `mash_playbook_reverse_proxyable_services_additional_labels` and `mash_playbook_reverse_proxyable_authenticationless_services_additional_labels` into them

PR #1052, opened 2025-10-03
Add Vikunja

luixxiul:
[Vikunja](https://vikunja.io/) is a self-hosted to-do app. Tested on my server and checked it worked, including the tag to create users. It should work with LLDAP to provide [LDAP authentication](https://vikunja.io/docs/ldap/) by setting [these environment variables](https://vikunja.io/docs/config-options/#1-auth-ldap) additionally.

### DIFF
--- a/docs/services/vikunja.md
+++ b/docs/services/vikunja.md
@@ -1,0 +1,263 @@
+<!--
+SPDX-FileCopyrightText: 2020 - 2024 MDAD project contributors
+SPDX-FileCopyrightText: 2020 - 2024 Slavi Pantaleev
+SPDX-FileCopyrightText: 2020 Aaron Raimist
+SPDX-FileCopyrightText: 2020 Chris van Dijk
+SPDX-FileCopyrightText: 2020 Dominik Zajac
+SPDX-FileCopyrightText: 2020 Mickaël Cornière
+SPDX-FileCopyrightText: 2022 François Darveau
+SPDX-FileCopyrightText: 2022 Julian Foad
+SPDX-FileCopyrightText: 2022 Warren Bailey
+SPDX-FileCopyrightText: 2023 Antonis Christofides
+SPDX-FileCopyrightText: 2023 Felix Stupp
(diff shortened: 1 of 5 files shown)


docs/services/vikunja.md
```diff
@@ -0,0 +1,259 @@
...
+# /vikunja                                                             #
+#                                                                      #
+########################################################################
+```
+
+**Note**: hosting Vikunja under a subpath (by configuring the `vikunja_path_prefix` variable) does not seem to be possible due to Vikunja's technical limitations.
+
+### Select database to use (optional)
+
+By default Vikunja is configured to use Postgres, but you can choose other databases such as MySQL (MariaDB) and SQLite. See [this section](https://github.com/mother-of-all-self-hosting/ansible-role-vikunja/blob/main/docs/configuring-vikunja.md#specify-database-optional) on the role's documentation for details.
```
spantaleev, line 66:
```suggestion
By default Vikunja is configured to use [Postgres](postgres.md) (if enabled), but you can choose other databases such as MySQL (MariaDB) and SQLite. If Postgres is not enabled, SQLite will be used. See [this section](https://github.com/mother-of-all-self-hosting/ansible-role-vikunja/blob/main/docs/configuring-vikunja.md#specify-database-optional) on the role's documentation for details.
```

^^ Repeating some stuff from the Requirements section, so that it's more obvious for people reading this one.

---------

On a related note, now I'm thinking that such automatic wiring and fallback could be dangerous.

Imagine not having Postgres enabled yet, because you're only hosting something like Vikunja. Vikunja will default to using SQLite. A few days later, you enable Postgres (because you need it for something else). Vikunja will now be auto-configured to use Postgres and your old existing Vikunja data will seem to disappear.

This is not specific to this role, I guess. We've likely done something similar for many others. Perhaps explicitly specifying the database type would solve it (the playbook could auto-wire Postgres only if `vikunja_database_type: postgres` and `postgres_enabled`).

luixxiul: Created a tracking issue here: https://github.com/mother-of-all-self-hosting/mash-playbook/issues/1061

PR #1324, opened 2025-11-19
Added Apache NiFi

spatterIight:
This is ready.

This container was not developed with IaC in mind, and does not follow Docker best practices.

In my testing running it as a non-root user was unsuccessful. The same with read-only mode.

Additionally, [configuration files ](https://github.com/spatterIight/ansible-role-nifi/tree/main/files/conf) and [scripts](https://github.com/spatterIight/ansible-role-nifi/tree/main/files/scripts) had to be hard-coded in the role for sane and consistent configuration methods. 

The scripts had to be modified from their [upstream versions](https://github.com/apache/nifi/tree/main/nifi-docker/dockerhub/sh) due to this ["bug"](https://issues.apache.org/jira/browse/NIFI-7060) which results in custom `nifi.properties` being overwritten with null values.

In the future, when the image is updated, it is likely new versions of the default configuration files will need to be committed to the role, and the same with patched versions of the newer scripts -- which is highly unfortunate.

Besides that it works great so far in my testing

### DIFF
--- a/docs/services/nifi.md
+++ b/docs/services/nifi.md
@@ -47,14 +47,11 @@ nifi_self_signed_cert_passphrase: ""
 # Generate one using `pwgen -s 64 1`, or some other way
 nifi_sensitive_props_key: ""
 
-# The admin username and password. The password must be 12 characters. The salt must be exactly 22 characters, and does not necessarily need to be changed from the example below.
-# Generate a password using `pwgen -s 32 1`, or some other way
-nifi_conf_login_identity_providers_xml:
-  - xpath: /loginIdentityProviders/provider/property[@name='Username']
-    value: admin
-
-  - xpath: /loginIdentityProviders/provider/property[@name='Password']
-    value: "{{ 'my-secure-password' | password_hash(hashtype='bcrypt', salt='0tfETQ5kFWhcT4oPwdbHbL') }}"
+# The default login credentials to configure
+# The password must be at least 12 characters, generate one using `pwgen -s 32 1`, or some other way
+# The salt must be exactly 22 characters, and does not necessarily need to be changed from default value
+nifi_login_username: admin
+nifi_login_password: my-secure-password
 
 ########################################################################
 #                                                                      #
@@ -71,7 +68,7 @@ nifi_conf_login_identity_providers_xml:
 traefik_provider_configuration_extension_yaml: |
   http:
     serversTransports:
-      insecure-nifi-transport:
+      {{ nifi_traefik_serverstransport }}:
         insecureSkipVerify: true
 
 ########################################################################
--- a/templates/requirements.yml
+++ b/templates/requirements.yml
@@ -477,7 +477,7 @@
   name: nextcloud
   activation_prefix: nextcloud_
 - src: git+https://github.com/spatterIight/ansible-role-nifi.git
-  version: v2.6.0-3
+  version: v2.6.0-4
   name: nifi
   activation_prefix: nifi_
 - src: git+https://seed.radicle.garden/z8vyzN3a8DmwhcUq3949SihKd1Wh.git


spatterIight:
This should be squashed on merge

spatterIight:
Hm actually I just looked at whats actually in toolkit.sh, and its literally nothing from what that document describes as being in the "admin toolkit".

They must be referring to another file.

None of these (original) scripts useful, since they are all related to modifying configuration files are run-time. Which we don't want because it causes conflicts with Ansible configuring those same files. Therefor I will tweak the role to shadow the directory, good catch!

docs/services/nifi.md
```diff
@@ -0,0 +1,99 @@
...
+
+- a [Traefik](traefik.md) reverse-proxy server
+
+## Prerequisites
+
+To deploy Apache NiFi using this role it is necessary that:
+
+1. The [community.general](https://github.com/ansible-collections/community.general) collection be installed. This is needed to support modifying XML configuration files.
+2. The [community.crypto](https://github.com/ansible-collections/community.crypto) collection be installed. This is needed to create the self-signed HTTPS certificate for Apache NiFi.
+3. The `keytool` program be available on the target host. This can be installed via `apt install default-jre` on Debian systems.
```
spantaleev: I suppose it's possible to run these tools in containers and avoid these dependencies.

In some roles we run `htpasswd` from some `httpd` container for the same reason.

For SSL certificates, an alternative may be to ask the user to generate the SSL certificate themselves and to provide it via `vars.yml`. Then Ansible would just install it. It makes things a bit harder on the user, but the final installation becomes more reproducible. Maybe it's not ideal, given that it will be harder on users, making them do manual manual work (which is better done by the tool).

docs/services/nifi.md
```diff
@@ -0,0 +1,99 @@
...
+# Generate one using `pwgen -s 64 1`, or some other way
+nifi_self_signed_cert_passphrase: ""
+
+# A passphrase used to encrypt sensitive values inputted into NiFi
+# Generate one using `pwgen -s 64 1`, or some other way
+nifi_sensitive_props_key: ""
+
+# The admin username and password. The password must be 12 characters. The salt must be exactly 22 characters, and does not necessarily need to be changed from the example below.
+# Generate a password using `pwgen -s 32 1`, or some other way
+nifi_conf_login_identity_providers_xml:
```
spantaleev: Why `_conf`?

If this affects some configuration setting for the installed component, the common convention in other roles is to use `_config`.

In this case, it seems like this configures the Ansible tooling with the replacements it needs to do.

Maybe `nifi_login_identity_providers_xml_replacements` is a good name.

We could split this into 3:

```yml
nifi_login_identity_providers_xml_replacements: "{{ nifi_login_identity_providers_xml_replacements_default + nifi_login_identity_providers_xml_replacements_auto + nifi_login_identity_providers_xml_replacements_custom }}"

nifi_login_identity_providers_xml_replacements_default:
  - xpath: /loginIdentityProviders/provider/property[@name='Username']
    value: "{{ nifi_login_username }}"

  - xpath: /loginIdentityProviders/provider/property[@name='Password']
    value: "{{ nifi_login_password | password_hash(hashtype='bcrypt', salt='???') }}"

nifi_login_identity_providers_xml_replacements_auto: []
nifi_login_identity_providers_xml_replacements_custom: []

nifi_login_username: admin
nifi_login_password: ''
```

This also moves the admin username and password into dedicated variables. `nifi_login_password` can be checked for non-emptiness in `tasks/validate_config.yml`.

-------

Maybe we need a variable for the salt too?

--------

On the other hand, hashing the password on each Ansible run (and getting a different hash) means that the file is being changed each time. This is not ideal.

This can be made more idempotent, by asking the user to do the bcrypt hashing themselves, outside of Ansible.
They can then provide it via a new variable like `nifi_login_password_hash`.

docs/services/nifi.md
```diff
@@ -0,0 +1,99 @@
...
+########################################################################
+#                                                                      #
+# traefik                                                              #
+#                                                                      #
+########################################################################
+
+traefik_provider_configuration_extension_yaml: |
+  http:
+    serversTransports:
+      insecure-nifi-transport:
```
spantaleev: Is this `insecure-nifi-transport` value also hardcoded somewhere in the NiFi Ansible role?

If so, it's probably better to extract it into a variable and reference it via the variable name here.

spatterIight: Will do! 

I originally had this as a variable but moved away from it since my Traefik instance is managed by Matrix -- and so the MASH NiFi role which defines this variable wasn't run and so the variable did not exist.

It is probably better to have it as a variable though, and have some comment about the above situation and what do to in it. I will make this change.

docs/services/nifi.md
spatterIight: The `_conf` name comes from the location of the files, in the role they are stored in the role at `files/conf` and inside the container in `/opt/nifi/nifi-current/conf`

I think your proposed variable pattern, `nifi_login_identity_providers_xml_replacements`, is fine. What is the `_auto` variable for in this case?

Also, doesn't the final variable need to do something like [this](https://github.com/spatterIight/ansible-role-nifi/blob/e1172b6876a2b16860937fec163ee7c9f1896be0/defaults/main.yml#L261) to allow overriding defaults?

--- 

> On the other hand, hashing the password on each Ansible run (and getting a different hash) means that the file is being changed each time. This is not ideal.

If the same salt is provided the hash is the same, if the salt is not provided idempotency is lost. Since the hash is the same if the salt is the same, I think supporting a new variable like `nifi_login_password_hash` is unnecessary.

PR #1354, opened 2025-11-23
Headplane fixup

spatterIight:
This MR adds a piece documentation that was missing, regarding a required variable. Additionally it updates the role itself. The role was updated to resolve a bug reported by QED.


spatterIight:
@QEDeD Can you review this to see if it fixes your issue

spatterIight:
This can be merged, I was able to replicate QED's issue and the fix resolves it, so probably ok to move forward w/o his feedback

PR #1384, opened 2025-11-29
Add Moodist

luixxiul:
[Moodist](https://moodist.mvze.net) is a free web application to mix and play ambient soundtracks (river, rain, cafe, airport, etc).

As using a root user is not really worth for running the streaming service, the role by default uses its own [Dockerfile](https://app.radicle.xyz/nodes/seed.radicle.garden/rad%3Az4Ez9aU119yFMs59cnA3KUnExFAeN/tree/51bdebb3210e646fb7e83666b8208d5a820ae0c5/templates/Dockerfile.j2) to replace `nginx` with `http-server`.

### DIFF
--- a/docs/services/moodist.md
+++ b/docs/services/moodist.md
@@ -59,6 +59,19 @@ moodist_hostname: moodist.example.com
 
 **Note**: hosting Moodist under a subpath (by configuring the `moodist_path_prefix` variable) does not seem to be possible due to Moodist's technical limitations.
 
+### Using the default Docker image (optional)
+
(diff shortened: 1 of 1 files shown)


luixxiul:
For BentoPDF:

```
IMAGE                              ID             DISK USAGE   CONTENT SIZE
alam00000/bentopdf-nonroot:v1.9.0  01b57b29180a        401MB         99.4MB 
alam00000/bentopdf-nonroot:v1.9.0  2b866ec2365e       76.5MB         19.9MB
```

It's massive reduction! :smile:

PR #1434, opened 2025-12-07
Add Mailcatcher

luixxiul:
[MailCatcher](https://mailcatcher.me) is the SMTP server which catches any message sent to it and displays in a web interface instead of sending it to the outside of the internal network.

I'm looking at https://github.com/axllent/mailpit as well; its optional POP3 server should make it possible to avoid using a third party mail service.

The authors claim that Mailcatcher has inspired [MailHog](https://github.com/mailhog/MailHog/blob/master/README.md), and MailHog has inspired [Mailpit](https://github.com/axllent/mailpit/blob/develop/README.md). Everything seems to be related in here.

### DIFF
--- a/docs/services/mailcatcher.md
+++ b/docs/services/mailcatcher.md
@@ -19,14 +19,14 @@ SPDX-License-Identifier: AGPL-3.0-or-later
 
 # MailCatcher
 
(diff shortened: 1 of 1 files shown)


luixxiul:
Having variety here does not seem to be a bad idea ;)

Issue #1539, opened 2025-12-26
Nextcloud - CODE websocket issue with traefik v3.6.4 - Encoded character rejection


LeBueth:
Hi,

Looks like the new rejection rule for encoded characters within traefik 3.6.4 (see [Traefik documentation](https://doc.traefik.io/traefik/migrate/v3/#v364) ) is responsible for the separate CODE and Nextcloud containers no longer working together. Configuration seems fine at first, the Office administration entry in Nextcloud gives a green light for connection in between CODE and Nextcloud. The Collabora-instance shows no problems as well, neither through the status page (`https://collabora.example.com` returning OK) nor through the admin interface (`https://collabora.example.com/browser/dist/admin/admin.html`). But when you're trying to edit an office document with Collabora, you'll face the issue. I always got a websocket connection error when trying to edit a document (odt) online, mentioning that it might be related to a faulty reverse proxy configuration. In non-debug-mode you can only see traefik throwing some 499s and 400s when trying to access the document. The logs revealed the problem after switching to DEBUG level on traefik:

> DBG github.com/traefik/traefik/v3/pkg/server/router/deny.go:52 > Rejecting request because it contains encoded character %3F in the URL path
> DBG github.com/traefik/traefik/v3/pkg/server/router/deny.go:52 > Rejecting request because it contains encoded character %2F in the URL path

A quick fix is to add the lines

```
traefik_config_entrypoint_web_secure_http_encodedCharacters_enabled: true
traefik_config_entrypoint_web_secure_http_encodedCharacters_allowEncodedQuestionMark: true
traefik_config_entrypoint_web_secure_http_encodedCharacters_allowEncodedSlash: true
```

to your vars.yml file.

Since I'm managing traefik with the matrix playbook, I'm wondering if there's a more elegant way to fix this? Maybe only for the communication in between Nextcloud and CODE? My guess would be to have a separate network spanning in between those two with the encodings allowed!? Haven't checked yet, if that's an option.

Version information:
Collabora Online: 25.04.7.2.1
Nextcloud: 32.0.2
Traefik: v3.6.4

luixxiul:
Fixed by https://github.com/mother-of-all-self-hosting/mash-playbook/pull/1543

PR #1543, opened 2025-12-28
Allow encoded characters in traefik for collabora online

zenphonix:
Traefik v3.6.4 introduced a more restrictive way of handling encoded characters in paths:
https://doc.traefik.io/traefik/migrate/v3/#v364
Which lead to an inaccessible Collabora Online websocket. Adding allowEncodedSlash and allowEndodedQuestionMark options in traefik fixes this issue.


zenphonix:
Fixes #1539

PR #2030, opened 2026-02-28
Wire Collabora restart_necessary to role-computed flag

QEDeD:
## Summary
This fixes a wiring bug in the generated service-manager list for Collabora.

- replace hardcoded `restart_necessary: true` with `restart_necessary: (collabora_online_restart_necessary | bool)`

## Why
The Collabora role already computes whether restart is necessary. Hardcoding `true` forces restart every run and conflicts with role restart semantics.

## Scope
- `templates/group_vars_mash_servers` only

## Validation
- confirmed rendered wiring now matches the role-computed restart flag logic


luixxiul:
@QEDeD does this PR include your configurations, doesn't it? It is not clear what is going on here.. 

Also it is required to update ansible-role-collabora-online itself to make `collabora_online_restart_necessary` effective; I will take care of it, thanks for the ping anyway!

PR #1684, opened 2026-01-21
Add OmniTools service

daguwar:
Adding the service [OmniTools](https://github.com/iib0011/omni-tools) which is a self-hosted collection of powerful web-based tools for everyday tasks.

PR #2364, opened 2026-04-14
Add LibreBooking

shukon:
Adding [LibreBooking](https://github.com/LibreBooking/librebooking) to MASH.

Used LLM (Claude) for style support to ensure a coherent style in tune with other roles.

Feel free to fork the role-repo into the MASH-organization if you prefer that.

(will only check on this on week days from ~ 09 to 13 since I'm doing this @ dayjob)

### DIFF
--- a/docs/services/librebooking.md
+++ b/docs/services/librebooking.md
@@ -79,15 +79,27 @@ The playbook automatically provisions a MariaDB database and user for LibreBooki
 
 ### First-time setup
 
-After running the playbook for the first time, navigate to:
 
+LibreBooking does **not** initialize its database schema automatically. You must run the web-based install wizard on first install.
+
+After running the playbook for the first time, retrieve your database credentials:
+
+```bash
+ansible-playbook -i inventory/hosts setup.yml --tags=print-db-credentials-librebooking
 ```
+
+Then navigate to the install wizard:
+
+```text
 https://booking.example.com/Web/install/
 ```
 
-Enter the `librebooking_install_password` you configured when prompted, then follow the wizard to initialize the database schema and create your administrator account.
+You will be prompted for the **installation password** you set above. On the next page:
 
-Keep `librebooking_install_password` set to a strong value to prevent unauthorized access to the setup page.
+- Enter the database credentials printed above as the **MySQL User** and **Password**
+- **Check "Import sample data"** — this creates the database schema and an initial `admin`/`password` account
+- Do **not** check "Create the database" or "Create the database user" — both already exist
+- Afterwards, log in with `admin`/`password` (and change the password)
 
 ## Usage
 
--- a/docs/services/librebooking.md
+++ b/docs/services/librebooking.md
@@ -79,7 +79,6 @@ The playbook automatically provisions a MariaDB database and user for LibreBooki
 
 ### First-time setup
 
-
 LibreBooking does **not** initialize its database schema automatically. You must run the web-based install wizard on first install.
 
 After running the playbook for the first time, retrieve your database credentials:
--- a/docs/services/librebooking.md
+++ b/docs/services/librebooking.md
@@ -48,23 +48,23 @@ librebooking_enabled: true
 
 librebooking_hostname: booking.example.com
 
-# To serve at a subpath instead of /, uncomment:
-# librebooking_path_prefix: /booking
-
 # Protects the /Web/install/ setup wizard. Use a strong password.
-librebooking_install_password: "your-strong-install-password-here"
+librebooking_environment_variables_lb_install_password: "your-strong-install-password-here"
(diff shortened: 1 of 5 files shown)


shukon:
I addressed the issues @luixxiul pointed out (thanks for that!).

templates/group_vars_mash_servers
```diff
@@ -9080,6 +9085,56 @@ languagetool_container_labels_traefik_tls_certResolver: "{{ traefik_certResolver
...
+########################################################################
+
+librebooking_enabled: false
+
+librebooking_identifier: "{{ mash_playbook_service_identifier_prefix }}librebooking"
+
+librebooking_base_path: "{{ mash_playbook_base_path }}/{{ mash_playbook_service_base_directory_name_prefix }}librebooking"
+
+librebooking_uid: "{{ mash_playbook_uid }}"
+librebooking_gid: "{{ mash_playbook_gid }}"
```
luixxiul: Doesn't it fail due to `/usr/local/bin/entrypoint.sh: line 50: pushd: /config: Permission denied` ?

shukon: It does. Must have slipped my tests because I edited manually I guess.
I ran it on a fresh system now, with your edits in the ansible-role too.
Changed to 33 as that is what is expected by the LibreBooking Dockerfile, do you have concerns about this?

luixxiul: Since you have confirmed that the change worked, I think it should be fine now.

templates/requirements.yml
```diff
@@ -432,6 +432,10 @@
   version: v6.7-1
   name: languagetool
   activation_prefix: languagetool_
+- src: git+https://github.com/shukon/ansible-role-librebooking.git
+  version: main
```
luixxiul: Would you please use a tag instead of `main` ? cf. https://github.com/mother-of-all-self-hosting/mash-playbook/pull/1684#issuecomment-3816392191

templates/group_vars_mash_servers
```diff
@@ -9080,6 +9085,56 @@ languagetool_container_labels_traefik_tls_certResolver: "{{ traefik_certResolver
...
+    +
+    ([mash_playbook_reverse_proxyable_services_additional_network] if mash_playbook_reverse_proxyable_services_additional_network else [])
+  }}
+
+librebooking_container_labels_traefik_enabled: "{{ mash_playbook_traefik_labels_enabled }}"
+librebooking_container_labels_traefik_docker_network: "{{ mash_playbook_reverse_proxyable_services_additional_network }}"
+
+# role-specific:mariadb
+librebooking_database_hostname: "{{ mariadb_connection_hostname if mariadb_enabled else '' }}"
+librebooking_database_password: "{{ (mash_playbook_generic_secret_key + ':db.' + librebooking_identifier) | hash('sha512') | to_uuid }}"
```
luixxiul:
```suggestion
librebooking_database_password: "{{ (mash_playbook_generic_secret_key + ':librebooking') | hash('sha512') | to_uuid }}"
```

templates/requirements.yml
shukon: Of course. Done.

luixxiul (approved):

docs/services/librebooking.md
```diff
@@ -0,0 +1,115 @@
...
+# Put a strong password below, generated with `pwgen -s 64 1` or another way
+mariadb_root_password: ""
+
+########################################################################
+#                                                                      #
+# /mariadb                                                             #
+#                                                                      #
+########################################################################
+
+
```
spantaleev: We don't typically quote the whole config (dependencies, etc.).

The fact that MariaDB is in **Dependencies** should be enough. People who need it, go to its dedicated docs page and copy its `vars.yml` snippet from there.

docs/services/librebooking.md
```diff
@@ -0,0 +1,115 @@
...
+#                                                                      #
+# /librebooking                                                        #
+#                                                                      #
+########################################################################
+```
+
+### Database
+
+The playbook automatically provisions a MariaDB database and user for LibreBooking and wires up the connection. You do not need to set database credentials manually.
+
```
spantaleev: This is also the case for all other services that depend on a database (MariaDB or Postgres).

While it's useful to new users (I guess), we don't typically repeat all this on each service's page. Hopefully, the rest of the documentation makes it clear.

docs/services/librebooking.md
shukon: Removed from the docs. Thanks.

docs/services/librebooking.md
shukon: Removed from the docs. Thanks.

PR #2451, opened 2026-04-25
Docs: move documentation out of examples/hosts and various clarifications

QEDeD:
I attempt to address the issue https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/4604, though I've ended up doing so for MASH rather than MDAD initially.

I've also added af few other changes that I believe would be helpful based on the issues/questions seen raised by users :)

### DIFF
--- a/docs/configuring-playbook.md
+++ b/docs/configuring-playbook.md
@@ -1,17 +1,17 @@
 <!--
-SPDX-FileCopyrightText: 2018 - 2025 Slavi Pantaleev
-SPDX-FileCopyrightText: 2018 - 2024 MDAD project contributors
+SPDX-FileCopyrightText: 2018-2024 MDAD project contributors
+SPDX-FileCopyrightText: 2018-2025 Slavi Pantaleev
 SPDX-FileCopyrightText: 2020 Sabine Laszakovits
 SPDX-FileCopyrightText: 2021 Cody Neiman
 SPDX-FileCopyrightText: 2021 Matthew Cengia
 SPDX-FileCopyrightText: 2021 Toni Spets
 SPDX-FileCopyrightText: 2022 Julian Foad
 SPDX-FileCopyrightText: 2022 Vladimir Panteleev
-SPDX-FileCopyrightText: 2022 - 2023 Julian-Samuel Gebühr
-SPDX-FileCopyrightText: 2023 MASH project contributors
-SPDX-FileCopyrightText: 2023 Shreyas Ajjarapu
+SPDX-FileCopyrightText: 2022, 2023 Julian-Samuel Gebühr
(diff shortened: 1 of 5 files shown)


docs/configuring-playbook.md
```diff
@@ -34,7 +34,54 @@ If you've configured your DNS records and retrieved the playbook's source code t
...
+
+If your SSH private key is protected by a passphrase, unlock it before running the playbook, for example with `ssh-agent` and `ssh-add`. Ansible's default SSH connection plugin does not provide an interactive channel for decrypting SSH private keys during a playbook run, so an encrypted key that has not been loaded may fail instead of prompting you. See Ansible's [SSH connection plugin documentation](https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/ssh_connection.html) for details.
+
+For encrypted Ansible Vault files, add `--ask-vault-pass` to your commands. Inventories using named or multiple Vault IDs may need `--vault-id <id>@prompt` instead. See Ansible's [Vault password documentation](https://docs.ansible.com/projects/ansible/latest/vault_guide/vault_managing_passwords.html). When both Vault and sudo passwords are required, pass both flags, for example:
+
+```sh
+just install-all --ask-vault-pass --ask-become-pass
```
luixxiul, line 70: As the documentation until this page has not yet mentioned how to use `just` (though mentioned on prerequisites.md, which is technically not true), I'd stick to the orthodox `ansible-playbook -i inventory/hosts setup.yml` command to be conservative.

QEDeD: That's a good point. 

I see three ways to handle this

1. Don't mention `just` at all, just stick with `ansible-playbook -i inventory/hosts setup.yml --ask-vault-pass --ask-become-pass`

2. Mention both. "plain" Ansible version and the just version. The pro of doing this is that it could be beneficial to quickly establish idea that `just` commands are generally used in place of `ansible-playbook` and the con is slightly higher complexity.

----------
```sh
ansible-playbook -i inventory/hosts setup.yml --ask-vault-pass --ask-become-pass
```
or the equivalent [just command](https://github.com/mother-of-all-self-hosting/mash-playbook/blob/main/docs/just.md):
```sh
just install-all --ask-vault-pass --ask-become-pass
```
----------

I think option 2 in which both get mentioned briefly strikes a good balance. What do you think? :)

luixxiul: I like the second option, as long as the command is corrected:

```sh
ansible-playbook -i inventory/hosts setup.yml --tags=install-all,start --ask-vault-pass --ask-become-pass
```

QEDeD: The PR should now be updated according to what we dicussed

PR #2884, opened 2026-06-23
Added vector

spatterIight:
Added vector


spatterIight:
This is ready. 

It should be a good alternative option to **promtail** which is currently end-of-life but still widely used in the industry (I believe)

PR #2903, opened 2026-06-26
Document how to enable sheets in cryptpad

microchipster:
Took me a little while to figure this out, so I thought it would be nice to add it explicitly in the docs in case it helps others out too.


luixxiul:
It was adjusted and added with https://github.com/mother-of-all-self-hosting/mash-playbook/commit/a5b36ac1c35d9c231efdffb42f1a4fd9a010aa4d. Thank you!

PR #3204, opened 2026-08-05
feat(ci): support all-hosts rollout for deploy/rollback setup-all

oliverlorenz:
## Summary
- `host` is now optional on `deploy-setup-all.yml` and `rollback-setup-all.yml` -- leaving it empty targets every host in the `mash_servers` inventory group instead of a single one.
- Empty host resolves the group via `ansible-inventory` in a new `resolve-hosts` job, then fans deploy/record/rollback out as a parallel matrix (`fail-fast: false` so one host's failure doesn't cancel the others).
- `deploy-run-setup-all.yml` gained a `rollback` input so each matrixed host resolves its own `deployed/<host>` tag independently, instead of relying on a shared upstream job's output (which GitHub Actions can't fan out per-matrix-instance to a downstream job).
- Dropped `record-rollback-success` from `deploy-setup-all.yml`: rollback redeploys exactly the sha its tag already points to, so re-tagging afterward was already a no-op.

## Test plan
- [x] `pre-commit run` (yamllint, ansible-lint, sync-deploy-workflow-options) passes on all changed files
- [x] YAML parses cleanly for all three workflow files
- [ ] Manual dry run: trigger `Deploy - setup-all` with a single `host` to confirm unchanged single-host behavior
- [ ] Manual dry run: trigger `Deploy - setup-all` with `host` empty to confirm the matrix fans out to all `mash_servers` hosts in parallel
- [ ] Manual dry run: force a failure and confirm the approval-gated auto-rollback rolls back every targeted host

🤖 Generated with [Claude Code](https://claude.com/claude-code)


oliverlorenz:
Opened by mistake against upstream instead of my fork -- closing. Sorry

